Issue #18, opened 2016-01-21
Gamejolt client windows: nw-gyp compile error.


luckcolors:
Hello again.
This time i'm tring to build the desktop app but of course it throws an error.
Here's the console ouput (apparently there isn't a way to make gulp more verbose so i'm sorry if there's some lack of info):
Command(all dependencies where installed): gulp --production --client --arch=32
[error.log.zip](https://github.com/gamejolt/gamejolt/files/99852/error.log.zip)
Note: Mine is a clean windows install. 
Also i have set the variable for gyp compiler as follows: GYP_MSVS_VERSION=2015

addaleax:
@luckcolors This is essentially nwjs/nw.js#4033, with a PR available bug not merged at nwjs/nw-gyp#81.

For the project maintainers: I just saw that you inserted some gulp task for re-compiling lzma-native for nw.js. Since v1.0.0, I use node-pre-gyp to provide pre-compiled binaries; I think I can look into supporting nw.js for that (so you don’t need to re-compile anything in most cases). Would you be interested?

hworld:
We could do that @addaleax, but I wouldn't want to give your more work than you'd need. nwjs is kind of a nightmare with compiled libs. I think it needs to be done per nwjs version, so it's probably not worth it to push precompiled. We do the compilation during the build here: https://github.com/gamejolt/gamejolt/blob/b38405c5ad672444c7d78f69ac50b1b71840237c/tasks/client.js#L181

It gets even crazier on Windows because you have to patch the compiled lib if you want to have an executable name for your application other than nw.exe. I guess they're fixing this on the new version that's in beta. We do that here in our build process: https://github.com/gamejolt/gamejolt/blob/b38405c5ad672444c7d78f69ac50b1b71840237c/tasks/client.js#L185

@luckcolors You can do a build using the command: `npm run client-build-win`. This calls the gulp build with all the correct options. However, currently it's exactly what you're calling, anyway. But in the future it'll make sure that you're in sync with what I run myself.

 Either way, I think I know what's happening. In your logs I see this:

```
gyp ERR! command "D:\\DevTools\\nodejs\\node.exe" "D:\\DevTools\\nodejs\\node_modules\\nw-gyp\\bin\\nw-gyp.js" "clean" "configure" "build" "--target=0.12.3" "--arch=ia32"
gyp ERR! cwd C:\Users\luckcolors\Desktop\gamejolt\build\client\prod\source\node_modules\lzma-native
[...]

PR #20, opened 2016-01-22
Update re-compilation script to use node-pre-gyp

addaleax:
Use node-pre-gyp to re-compile lzma-native if necessary. Using the `install` subcommand instead of `build` makes node-pre-gyp look for precompiled binaries for nw.js, which are available as of `lzma-native@1.0.4`. (This is not logically dependent on gamejolt/client-voodoo#15, but enabling pre-built binaries for nw.js seems like a cool thing! :))

See also #18.

### DIFF
--- a/tasks/client.js
+++ b/tasks/client.js
@@ -45,7 +45,7 @@ module.exports = function( config )
 	}
 
 	gulp.task( 'client:gyp', shell.task( [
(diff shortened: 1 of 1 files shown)


addaleax:
Obviously closing in favour of #22.

PR #22, opened 2016-01-27
Use lzma-native with node-pre-gyp

YLivay:
When merging remove the git reference. It needs the changes in gamejolt/client-voodoo#update-lzma-dep pulled in first


addaleax:
@YLivay I’d still suggest you use `node-pre-gyp clean install` instead of `node-pre-gyp clean configure build`, so that the build for nw.js does not require compilation, but in the end it’s up to you whether you always want to re-compile or not.

hworld:
Would love to be able to just do install. For some reason the build failed on windows with "can't find nw.exe" for some reason. nw.exe doesn't work for some reason, nw does... it may be an nvm path issue or something...

addaleax:
Hm, weird – `node-pre-gyp install` should not call any kind of node or nw.js process. Sorry for all the trouble, but could you maybe upload your npm-debug.log (or equivalent) here?

PR #24, opened 2016-02-04
Muting users through messages.

Agrosis:


src/app/components/chat/window/window.html
```diff
@@ -164,8 +164,12 @@ <h3 ng-if="room.type == Chat_Room.TYPE_PM" class="anim-fade-in-right no-animate-
 							{{ ::msg.loggedOn | amDateFormat:'h:mm A' }}
 						</span>
 
-						<div class="chat-window-message-content-wrap">
-							<div ng-if="::msg.contentRaw.length < 500" class="chat-window-message-content highlight" ng-bind-html="::msg.content"></div>
+						<div ng-if="::msg.contentRaw.length < 500" class="chat-window-message-content-wrap">
+							<div class="chat-window-message-content highlight" ng-bind-html="::msg.content"></div>
+
+							<div class="chat-window-message-mod-tools" ng-if="client.canModerate( room, msg.user )">
```
hworld: bind-once this with `::` since it will never change after it's rendered

src/app/components/chat/window/window.html
```diff
@@ -164,8 +164,12 @@ <h3 ng-if="room.type == Chat_Room.TYPE_PM" class="anim-fade-in-right no-animate-
...
+								<span class="jolticon jolticon-friend-remove-2" tooltip="Mute User" ng-click="ctrl.muteUser( msg.user.id )"></span>
```
hworld: I feel like this should be in an `<a>` tag with ng-click on it. Also put the tooltip on the A tag and change to using `gj-tooltip` instead for this case.

src/app/components/chat/window/window.html
```diff
@@ -180,8 +184,11 @@ <h3 ng-if="room.type == Chat_Room.TYPE_PM" class="anim-fade-in-right no-animate-
 							<p ng-if="msg._collapsable && !msg._expanded">
 								<a class="hidden-text-expander" ng-click="msg._expanded = true"></a>
 							</p>
-						</div>
 
+							<div class="chat-window-message-mod-tools" ng-if="client.canModerate( room, msg.user )">
```
hworld: Same comments as above.

PR #79, opened 2016-12-30
Fold fireside

hworld:


hworld:
no longer relevant, will be handled by user posts

PR #100, opened 2017-09-12
Show user instead of target user.

bowenwaregames:
A small fix in regards to this issue:
https://github.com/gamejolt/issue-tracker/issues/964


hworld:
This won't work, either. You actually have to use the userFriendship.getThem method. You will have to pass in the current user to the method and it returns the correct user that's not "us".

hworld:
nvm, i'm dumb

PR #105, opened 2017-09-26
Turn off text shadow for hashtags

niksudan:
Closes gamejolt/issue-tracker#1002


niksudan:
Wasn't sure whether to have it as `a.hashtag` or just `.hashtag`. Let me know if you would rather target it as a sibling.

PR #109, opened 2017-10-28
New game page design

hworld:
It combines devlogs and game pages. Uses 2 columns going down the whole page instead of splitting it into rows. This allows the devlog posts to be higher up and infinite scrolling for all game pages.


hworld:
This took me almost a year. Geez

PR #171, opened 2017-12-13
hide vote option for locked topics and add archived sort mode

nilllzz:


src/app/components/forum/topic-list/topic-list.html
```diff
@@ -3,11 +3,11 @@
 	<div class="forum-topic-list-item"
 		v-for="topic of topics"
 		:key="topic.id"
-		:class="{ '-has-voting': useUpvotes }"
+		:class="{ '-has-voting': useUpvotes && !topic.is_locked }"
```
hworld: may as well use a getter in the component class: `shouldShowVoting` since we do this multiple times

src/app/views/forums/channels/view/view.html
```diff
@@ -111,6 +111,18 @@ <h1>
 							<translate>Top</translate>
 						</router-link>
 					</li>
+					<li v-if="channel.type === 'voting'">
```
hworld: decided to show archive tab for all channels, not just voting

PR #175, opened 2018-01-19
Anti spam

hworld:


src/auth/views/auth/join/join.ts
```diff
@@ -34,14 +35,17 @@ export default class RouteAuthJoin extends BaseRouteComponent {
 		return this.$gettext('auth.join.page_title');
 	}
 
-	async onJoin(formModel: any) {
+	async onJoining(formModel: FormModel) {
 		// We store these so we can log them in automatically once their
 		// verification happens.
 		this.setCredentials({
```
YLivay: Get rid of setCredentials through the store - pass these in through session storage, including the auth token. This way, we can even auto login people from the app section of the site.

src/auth/components/forms/twitter-email/twitter-email.html
```diff
@@ -26,6 +26,12 @@
 		>
 		auth.linked_account.twitter.finalize.duplicate_email_error_html
 	</div>
+	<div class="alert alert-warning anim-fade-in-enlarge no-animate-leave"
```
YLivay: Instead of reporting the error here - consider doing the rate limit at an earlier stage to make this the same flow as the other auth operations.

src/auth/views/auth/join/join.ts
YLivay: Also, don't emit join event. simply do the router push in this function.

src/auth/views/auth/join-almost/join-almost.html
```diff
@@ -1,25 +1,72 @@
 <div class="anim-fade-in-up">
-	<h2 class="section-header">
-		<translate>auth.join.almost.heading</translate>
-	</h2>
+	<template v-if="doingCaptcha">
```
YLivay: Split the captcha and authorize pages to `join/captcha` and `join/almost`

src/auth/views/auth/join-almost/join-almost.html
```diff
@@ -1,25 +1,72 @@
...
+			<button
+				class="btn"
+				@click="gotoJoin"
+				v-if="!serverErrors['rate-limit']"
+				>
+				<app-jolticon icon="reply" />
+				<translate>Retry?</translate>
+			</button>
+		</div>
+		<div v-translate>
```
YLivay: Change to <translate< wherever v-translate is not needed. We only need to use it when we have parameters or plural manipulations or inner html.

src/auth/views/auth/join-almost/join-almost.ts
```diff
@@ -8,21 +8,85 @@ import { AppProgressPoller } from '../../../../lib/gj-lib-client/components/prog
...
+
 	get routeTitle() {
 		return this.$gettext('auth.join.almost.page_title');
 	}
 
+	routeInit() {
+		this.doingCaptcha = !!this.$route.params.token;
+
+		this.recaptchaInterval = setInterval(() => {
+			if (grecaptcha) {
```
YLivay: Instead of recaptcha interval mess - make a component that is in charge of loading. Look at facebook/twitter/youtube sdks for reference.

PR #176, opened 2018-01-20
Grid

hworld:


src/app/components/activity/feed/notification/notification.html
```diff
@@ -9,7 +9,7 @@
 			</div>
 
 			<div class="timeline-list-item-title">
-				<app-notification-descriptive-action :notification="notification" />
+				{{ this.titleText }}
```
hworld: no need to use `this.` here in vue. it adds it by default

src/app/components/grid/client.ts
```diff
@@ -0,0 +1,126 @@
+import { Socket, Channel } from 'phoenix';
```
hworld: name this client.service.ts. we reserve the non-suffixed for vue components

src/app/components/grid/client.ts
```diff
@@ -0,0 +1,126 @@
...
+import { Growls } from '../../../lib/gj-lib-client/components/growls/growls.service';
+import { Environment } from '../../../lib/gj-lib-client/components/environment/environment.service';
+import * as nwGui from 'nw.gui';
+import { store } from '../../store/index';
+import {
+	Notification,
+	getNotificationText,
+} from '../../../lib/gj-lib-client/components/notification/notification-model';
+
+function getCookie(name: string): Promise<string | undefined> {
```
hworld: what do you think about putting this into a new file and exporting it? maybe in frontend-lib/components/cookie/cookie.service.ts or something? it can just export this one func, but it would be nice since we use it both here and in chat now

hworld: actually, maybe not frontend-lib since it depends on nw.gui which is included only in gamejolt repo

hworld: i did this

src/app/components/grid/client.ts
```diff
@@ -0,0 +1,90 @@
...
+			// not properly logged in
+			return;
+		}
+
+		console.log('Connecting to the Grid!');
+
+		const userId = user.id.toString();
+
+		this.socket = new Socket(Environment.gridHost);
+		this.socket.connect({
```
hworld: got this message: `passing params to connect is deprecated. Instead pass :params to the Socket constructor`

src/_common/cookie/cookie.service.ts
```diff
@@ -0,0 +1,37 @@
...
+				y = ARRcookies[i].substr(ARRcookies[i].indexOf('=') + 1);
+				x = x.replace(/^\s+|\s+$/g, '');
+				if (x === name) {
+					return resolve(decodeURI(y));
+				}
+			}
+			return resolve(undefined);
+		}
+	});
+}
```
YLivay: This looks like ass, thanks prettier.

nilllzz: np

hworld: ur welcome

PR #209, opened 2018-09-26
Reactive activity feed

nilllzz:


src/app/components/activity/feed/feed-container-service.ts
```diff
@@ -30,6 +30,9 @@ export interface ActivityFeedContainerOptions {
 	notificationWatermark?: number;
 }
 
+const SCROLL_DIRECTION_FROM = 'from';
```
hworld: Change to `ScrollDirectionFrom` casing

src/app/components/activity/feed/feed-container-service.ts
```diff
@@ -100,6 +104,11 @@ export class ActivityFeedContainer {
 		this.processGames();
 	}
 
+	clear() {
```
hworld: Gotta clear way more than this.

PR #221, opened 2018-10-25
New content editor

hworld:


gulpfile.js
```diff
@@ -67,6 +67,12 @@ const config = {
 			server: true,
 			crawl: true,
 		},
+		z: {
+			title: 'Game Jolt - Indie games for the love of it',
+			server: true,
```
hworld: this should be false (or left out if false is default), we don't need ssr

src/app/views/discover/games/view/view.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-	<div class="route-discover-games-view" v-if="game">
+	<div class="route-discover-games-view" v-if="game && game.developer">
```
YLivay: Why this check? when would it not have a developer?

nilllzz: Removed.

PR #231, opened 2018-11-14
Add fireside post tags suggestion

YLivay:


src/app/components/forms/game/description/tags/tags.ts
```diff
@@ -1,64 +1,41 @@
-import View from '!view!./tags.html?style=./tags.styl';
```
hworld: I don't think I like this refactor...
I think we should get rid of game-description-tags component and instead put the logic in the main form and then funnel off to the app-tag-suggestions component from there.
I don't think we should have the tag-suggestion component take a TagSuggester. I think it should basically take hte props that this component takes.
If you want, inside that component, you can still make a TagSuggester model, though.

src/app/components/forms/post/post.ts
```diff
@@ -201,6 +204,11 @@ export class FormPost extends BaseForm<FormPostModel>
 		return this.longEnabled;
 	}
 
+	get tagText() {
+		const longText = this.longEnabled ? this.formModel.content_markdown : '';
+		return `${this.formModel.lead} ${longText || ''}`.toLowerCase();
```
hworld: Maybe the tag-suggester component should lowercase, so that hte logic doens't need to be here.

src/app/components/forms/post/post.ts
```diff
@@ -415,6 +426,7 @@ export class FormPost extends BaseForm<FormPostModel>
 
 	onLoad(payload: any) {
 		this.keyGroups = KeyGroup.populate(payload.keyGroups);
+		this.possibleTags = payload.tags;
```
hworld: Maybe this should be called "featuredTags"?

src/app/components/forms/post/tags/tags.html
```diff
@@ -0,0 +1,15 @@
+<div v-if="shouldShow">
```
hworld: Same as game description, maybe just put this in the main component, now that tag-suggestion does most of the work.

src/app/components/forms/game/description/tags/tags.ts
YLivay: I don't like having too much styling in the tag suggestion component. It should be scoped down to the essentials - suggesting tags. The help text block and wether it should show at all should be handled by the parent using it, but since you cant do a v-if based on a component's value you're using inside the v-if i had to refactor it so that it'd be usable before the component even gets made. That's why tag suggester is a thing, mostly.

We should talk about this, im not sure i agree hmmm

src/app/components/forms/post/post.ts
YLivay: yes

PR #241, opened 2018-12-04
hide public rooms section when no rooms got returned

nilllzz:


src/app/components/chat/sidebar/sidebar.html
```diff
@@ -1,7 +1,9 @@
 <div class="chat-sidebar fill-darkest">
 
 	<!-- Public Chat Rooms -->
-	<div class="chat-rooms-list">
+	<div v-if="hasPublicRooms"
```
hworld: the v-if should be on its own line

PR #244, opened 2018-12-04
restrict post modal from opening multiple times

nilllzz:


hworld:
I don't think we ever want to show multiple post edit modals, so maybe we don't do all this block/unblock, and instead just set a modal ID that will be the same always, no matter the post being edited...

PR #253, opened 2019-01-08
Community Known Members

nilllzz:


src/app/views/communities/view/overview/overview.ts
```diff
@@ -101,11 +113,25 @@ export default class RouteCommunitiesViewOverview extends BaseRouteComponent {
 		return false;
 	}
 
+	get shouldShowKnownMembers() {
+		return this.app.user && this.knownMembers && this.knownMembers.length > 0;
```
YLivay: do we need to check for this.app.user?

src/app/views/communities/view/overview/overview.ts
```diff
@@ -62,10 +69,15 @@ export default class RouteCommunitiesViewOverview extends BaseRouteComponent {
 	@Prop(Array)
 	tags!: string[];
 
+	@State
+	app!: Store['app'];
```
YLivay: I dont think you need this.

nilllzz: I SAID WE NEED IT

PR #254, opened 2019-01-10
Community Unread Watermark

nilllzz:


hworld:
Pass this data as a prop into the child route instead of assigning onto community model.

PR #255, opened 2019-01-14
Better display of small images in posts.

hworld:


src/app/components/activity/feed/devlog-post/media/item/item.styl
```diff
@@ -0,0 +1,34 @@
...
+	position: relative
+	display: inline-block
+	vertical-align: middle
+	width: 100%
+	max-width: 100% !important
+
+// The "item" gets the correct dimensions applied, so we want to stretch
+// out any image or video in the item to be full width/height.
+.-img, .-video
+	rounded-corners()
```
YLivay: Why rounding corners?

hworld: it's what we do

PR #259, opened 2019-01-15
Make the dark mode the default.

hworld:


src/_common/settings/settings.service.ts
```diff
@@ -12,7 +12,7 @@ interface Setting {
 const defaultSettings: { [k: string]: Setting } = {
 	'theme-dark': {
 		type: 'boolean',
-		val: false,
+		val: true,
```
YLivay: change to !0 , ive seen the pros do it

PR #260, opened 2019-01-15
Removed a ton of ad code that we no longer need.

hworld:


src/app/components/shell/footer/footer.html
```diff
@@ -32,10 +32,6 @@
 		<br class="visible-xs">
 
 		<div class="clearfix">
-			<div class="footer-ad" v-if="!Screen.isXs && shouldShowAd">
```
YLivay: enjoy merging

PR #263, opened 2019-02-16
Fireside post moderate

nilllzz:


src/app/components/post/view/view.ts
```diff
@@ -60,7 +60,10 @@ export class AppPostView extends Vue {
 	}
 
 	get shouldShowManage() {
-		return this.post && this.post.isManageableByUser(this.app.user);
+		return (
+			(this.app.user && this.app.user.isMod) ||
+			(this.post && this.post.isManageableByUser(this.app.user))
```
YLivay: don't we have to check for ths.app.user in here too?
`this.app.user && (this.app.user.isMod || (this.post && this.post.isManageableByUser(this.app.user)))`
feel free to  break it up into more readable if statements, this is getting ridiculous

nilllzz: It wasn't there before, so I assume not

PR #264, opened 2019-02-16
implement new sketchfab model url schema

nilllzz:


src/app/components/forms/post/post.ts
```diff
@@ -111,8 +111,8 @@ export class FormPost extends BaseForm<FormPostModel>
 	};
 
 	readonly YOUTUBE_URL_REGEX = /^(https?:\/\/)?(www\.)?(youtube\.com\/watch\?v=|youtu\.be\/)([a-zA-Z0-9_\-]{11})(&.+)*$/i;
-	readonly SKETCHFAB_URL_REGEX = /^(https:\/\/)?(www.)?sketchfab.com\/models\/[0-9a-f]{32}\/?$/i;
-	readonly SKETCHFAB_FIELD_REGEX = /^((https:\/\/)?(www.)?(sketchfab.com\/models\/[0-9a-f]{32}\/?))|([0-9a-f]{32})$/i;
+	readonly SKETCHFAB_URL_REGEX = /^((https:\/\/)?(www.)?(sketchfab.com\/3d-models\/([a-z0-9]+-)+[0-9a-f]{32}\/?))\/?$/i;
```
YLivay: Use non capturing groups for anything you dont care to match on. Make sure the model id is the only bit that has a capturing group. Something like `3d-models\/[\w\d\-]+?-([0-9a-f]{32})\/?` seems to work better.

Make sure to use that matcher instead of substring later on to extract the model ID.

src/app/components/forms/post/post.ts
```diff
@@ -185,10 +185,12 @@ export class FormPost extends BaseForm<FormPostModel>
 	get sketchfabId() {
 		if (this.formModel.sketchfab_id.match(this.SKETCHFAB_URL_REGEX)) {
 			// extract model id from url
-			const matches = this.formModel.sketchfab_id.match(/[a-f0-9]{32}/i);
-			if (matches && matches.length > 0) {
-				return matches[0];
+			let url = this.formModel.sketchfab_id;
+			while (url.endsWith('/')) {
```
YLivay: Lets not do this. Make sure you can use the matches from the regex.

nilllzz:
Used non capture group and regex to get id

PR #273, opened 2019-04-03
Just some random misc fixes

YLivay:


src/app/components/client/install-package-modal/install-package-modal.ts
```diff
@@ -43,6 +43,17 @@ export default class AppClientInstallPackageModal extends BaseModal {
...
 
+	get installablePackages() {
+		const packagesById = arrayIndexBy(this.packageData.packages, 'id');
+		return Object.keys(this.buildsByPackage)
+			.map(pkgId => packagesById[pkgId])
+			.filter(pkg => !!pkg);
+
+		// Can probably also just do this, but the above seems more resillient.
+		// I don't want to assume installableBuilds is always necessarily computed from this.packageData.packages
+		// return this.packageData.packages.filter(p => !!this.buildsByPackage[p.id]);
```
hworld: This is a weird comment. Did you want to just explain why do it with object.keys/map/filter above the line that does it?

YLivay: yes. is it weird?

PR #286, opened 2019-06-03
Add user follow popover when liking posts

YLivay:
Also, the new and existing user follow popover no longer show if they are dismissed a few times.


src/app/components/activity/feed/event-item/event-item.ts
```diff
@@ -164,14 +150,18 @@ export default class AppActivityFeedEventItem extends Vue {
...
-	get shouldShowFollow() {
+	get shouldShowFollowInHeader() {
 		// Don't show follow for game posts. Only for user posts/videos.
-		return (
-			this.feed.shouldShowFollow &&
-			!(this.post && this.post.game) &&
-			this.user &&
-			!this.user.is_following
-		);
+		if (!this.feed.shouldShowFollowInHeader || !this.post || this.post.game) {
```
hworld: You say videos, but then check if post exists. For comment videos, won't `this.post` be null?

src/app/components/activity/feed/event-item/event-item.vue
```diff
@@ -107,9 +107,10 @@
 				/>
 
 				<!--
-				This shouldn't ever really show a collapser. It's for the jokers that think it would
-				be fun to make a post with a bunch of new lines.
-			-->
+					This shouldn't ever really show a collapser. It's for the jokers that think it would
+					be fun to make a post with a bunch of new lines.
+					TODO: is item.isLeadOpen wrong?
```
hworld: figre this out before committing. I don't like TODO lines being left in

src/app/components/activity/feed/event-item/event-item.vue
```diff
@@ -152,7 +153,13 @@
 				/>
 			</div>
 
-			<app-event-item-controls class="-controls" :post="post" :video="video" @expand="onExpand()" />
+			<app-event-item-controls
+				class="-controls"
+				:post="post"
+				:video="video"
+				:show-user-follow-on-like="feed.shouldShowFollowOnLike"
```
hworld: shorter attr name would be appreciated

src/app/components/game/follow-widget/follow-widget.ts
```diff
@@ -90,14 +102,16 @@ export default class AppGameFollowWidget extends Vue {
 			// Do this before attempting to follow.
 			// We don't want to wait till the follow is confirmed to show the dialog,
 			// and even if the follow fails it's not like we'll close it.
-			if (this.shouldShowFollow && !this.game.developer.is_following) {
+			if (this.shouldShowFollow && UserFollowSuggestion.canSuggest(this.game.developer.id)) {
```
hworld: Shouldn't the canSuggest function call be put into the getter for `shouldShowFollow`? or is it because it won't trigger the change detection? If so, put a comment

src/app/components/game/follow-widget/follow-widget.ts
```diff
@@ -110,4 +124,10 @@ export default class AppGameFollowWidget extends Vue {
 			}
 		}
 	}
+
+	onFollowPopoverDismissed() {
+		if (!this.game.developer.is_following) {
+			UserFollowSuggestion.dontSuggest(this.game.developer.id);
```
hworld: I don't like using words that should have an apostrophe. Call `noSuggest` or `doNotSuggest` or something else

src/app/components/activity/feed/event-item/event-item.vue
YLivay: This is a todo to whoever wrote this, I don't understand where it was expecting to get this value. Can you or nils explain this?

src/app/components/activity/feed/event-item/event-item.ts
YLivay: I didnt touch the comment, i just split that single return into a more readable form haha.

I guess the original intention was to support follows on comment videos but that doesnt seem implemented hmm. For now i'll remove the bit about videos I guess.

src/app/components/activity/feed/event-item/event-item.vue
YLivay: Any less specific and its too ambiguous. if I don't say `show` it means it'll auto follow. If I don't say `User`, it'll be confused with following the resource the event item was posted for. e.g. the game a post was made on. if I remove `OnLike` you might assume it always shows, or you wouldn't know which control would show the follow widget. Hmm..

I guess `OnLike` is the most specific bit i can omit, so `showUserFollow`?

src/app/components/activity/feed/event-item/event-item.ts
hworld: I mean the code is wrong, not the comment. Shouldn't it be this?
`if (!this.feed.shouldShowFollowInHeader || (this.post && this.post.game)) {`

src/app/components/activity/feed/event-item/event-item.vue
hworld: yeah, I liked "show-user-follow" since the controls can figure out how it wants to do that

src/app/components/activity/feed/event-item/event-item.vue
nilllzz: I didn't write this, no idea what `isLeadOpen` is supposed to be. It was probably valid at some point and moved to some other part of the code.

hworld: We can talk about this on chat. It seems like I didn't properly port this one line over to the new "activity view" thing I did. Basically, it's supposed to now call into `isItemLeadOpen` in the `src/app/components/activity/feed/view.ts` class.

PR #287, opened 2019-07-02
More checkmarks

nilllzz:


src/app/views/profile/profile.vue
```diff
@@ -200,10 +194,6 @@
 </template>
 
 <style lang="stylus" scoped>
```
YLivay: remove this empty style block then

PR #292, opened 2019-07-09
Make search focus nicer in mobile

YLivay:


src/app/views/search/search.ts
```diff
@@ -71,4 +76,8 @@ export default class RouteSearch extends BaseRouteComponent {
 			!this.searchPayload.postsCount
 		);
 	}
+
+	mounted() {
+		this.$refs.search.focus();
```
hworld: Should app-search component have a prop called "autofocus" that just does this work when it is mounted? THis way you don't have to reach in with a ref on this page.

PR #294, opened 2019-07-15
Better client logs

YLivay:


src/app/components/client/system-report-modal/system-report-form.vue
```diff
@@ -1,5 +1,24 @@
 <template>
 	<app-form name="systemReportForm">
+		<app-form-group name="issue_tracker_url" optional :label="$gettext('Issue Tracker Link')">
+			<p class="help-block">
+				<translate>
+					Paste the link to your ticket here. It's optional, but without it we can't get back to you
+					if we need more info to solve it.
+				</translate>
+				<span class="emoji emoji-huh" />
```
nilllzz: Eh? No app-jolticon?

src/app/components/client/system-report-modal/system-report-form.vue
```diff
@@ -1,5 +1,24 @@
 <template>
 	<app-form name="systemReportForm">
+		<app-form-group name="issue_tracker_url" optional :label="$gettext('Issue Tracker Link')">
+			<p class="help-block">
+				<translate>
+					Paste the link to your ticket here. It's optional, but without it we can't get back to you
```
nilllzz: I think we should directly link to the issue tracker project here.

src/app/components/client/system-report-modal/system-report-form.vue
```diff
@@ -1,5 +1,24 @@
...
+					if we need more info to solve it.
+				</translate>
+				<span class="emoji emoji-huh" />
+			</p>
+
+			<app-form-control
+				type="text"
+				placeholder="https://github.com/gamejolt/issue-tracker/issues/1313"
+				:rules="{
+					max: 255,
```
nilllzz: Probably should also have a min length (of 42, which would be the length of this: `github.com/gamejolt/issue-tracker/issues/`), or a regex to test for that url shape.

src/app/components/client/system-report-modal/system-report-form.vue
YLivay: It's a game jolt emoji, not a jolticon. iirc we don't have our emojis as jolticons.

src/app/components/client/system-report-modal/system-report-form.vue
YLivay: We do in the modal that uses the form so its clear enough. It's also the only link there so it stands out too.

PR #297, opened 2019-07-23
Add margin to "X New Items" button in feed on mobile

nilllzz:
On mobile the button was directly attached to the top post in the feed, without any border or margin.

### DIFF
--- a/src/app/components/activity/feed/new-button/new-button.vue
+++ b/src/app/components/activity/feed/new-button/new-button.vue
@@ -9,7 +9,6 @@
 <style lang="stylus" scoped>
 @require '../variables'
 @require '~styles-lib/mixins'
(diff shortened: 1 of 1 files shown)


src/app/components/activity/feed/new-button/new-button.vue
```diff
@@ -9,10 +9,10 @@
 <style lang="stylus" scoped>
 @require '../variables'
 @require '~styles-lib/mixins'
+@require '~styles/variables'
```
hworld: this isn't needed, the first variables line imports it in

nilllzz: 👌

hworld:
deploy it

PR #301, opened 2019-07-26
Verification Application

nilllzz:


src/app/components/forms/verified-account/verified-account.vue
```diff
@@ -0,0 +1,21 @@
+<template>
+	<app-form name="verifiedAccountForm">
+		<app-form-group name="application" :label="$gettext(`Application Text`)">
+			<app-form-control-textarea :maxlength="maxChars" rows="2" :disabled="!canSubmit" />
+			<app-form-control-errors />
+			<div v-if="!existingApplication" class="help-block">
+				Provide us with a short explanation as to why you think your account should be verified!
```
YLivay: Assuming no translate block because we do require english for this section?

src/app/views/dashboard/account/account.vue
```diff
@@ -85,6 +85,14 @@
 										<translate>Password</translate>
 									</router-link>
 								</li>
+								<li v-if="shouldShowVerify">
+									<router-link
+										:to="{ name: 'dash.account.verified-account' }"
+										active-class="active"
+									>
+										<translate>Verify Account</translate>
```
YLivay: Couldn't this be confused with that annoying step or verifying one's account through the email link? How else can we call it to keep it short and clear?

src/app/components/forms/verified-account/verified-account.vue
nilllzz: Correct.

src/app/views/dashboard/account/account.vue
nilllzz: Maybe "Verification Application"?

PR #305, opened 2019-07-30
Inline comment buttons

nilllzz:


src/app/components/event-item/controls/controls.vue
```diff
@@ -56,6 +56,7 @@
 							v-else
 							resource="Fireside_Post"
 							:resource-id="post.id"
+							:editor-startup-arg="clickedCommentType"
```
YLivay: Change according to what we decide in frontend-lib

PR #306, opened 2019-07-30
Pinned posts

nilllzz:


src/app/components/activity/feed/event-item/event-item.ts
```diff
@@ -176,6 +176,14 @@ export default class AppActivityFeedEventItem extends Vue {
 		return this.post && this.post.isManageableByUser(this.app.user);
 	}
 
+	get isPinned() {
+		return (
+			this.post instanceof FiresidePost &&
+			!!this.post.is_pinned &&
+			this.feed.shouldShowEditControls
```
YLivay: why do we only consider a post pinned if we're showing the edit controls?

src/app/components/activity/feed/view.ts
```diff
@@ -149,6 +151,45 @@ export class ActivityFeedView {
 		}
 	}
 
+	processPinned(input: EventItem) {
```
YLivay: I don't understand this. let's talk about it.

src/app/components/activity/feed/view.ts
```diff
@@ -149,6 +151,45 @@ export class ActivityFeedView {
...
+				}
+
+				// Not a user post from the same user
+				if (!itemIsGame && !inputIsGame) {
+					if (input.action.user.id !== item.feedItem.action.user.id) {
+						return;
+					}
+				}
+
+				item.feedItem.action.is_pinned = false;
```
YLivay: I don't like this - the post is pinned, just not in the feed so setting the is_pinned to false might have unknown side effects if we use the same instance of the post in other places besides the feed. this should probably be some state on the feed item or a getter on the feed service.

src/app/components/fireside/post/manage/manage.vue
```diff
@@ -71,6 +71,17 @@
 						</template>
 
 						<div class="list-group list-group-dark">
+							<template v-if="shouldShowPin">
```
YLivay: I think we should show the pin controls, only disable it and have a tooltip explaining why you can't pin this specific post.

PR #308, opened 2019-08-08
Community description

nilllzz:


src/_common/content/content-document.ts
```diff
@@ -78,4 +78,17 @@ export class ContentDocument extends ContentNode {
 		// Don't count the first paragraph.
 		return super.getLength() - 1;
 	}
+
+	/**
+	 * Ensures that the last content object is a paragraph.
```
hworld: You should describe why this is needed. We need it because loading in the content editor is gonna add the node anyway, since it always requires the last node to be a paragraph.

nilllzz: I have a comment for it where it's used in content-editor.ts

nilllzz: done.

PR #310, opened 2019-08-12
Use favicon in ssr, update favicon to the recommended guidelines for Google

YLivay:


src/index-ssr.html
```diff
@@ -54,6 +54,9 @@
 		<!-- Twitter -->
 		<meta name="twitter:site" content="@gamejolt">
 
+		<!-- Favicon -->
+		<link rel="shortcut icon" type="image/png" href="/favicon.png"/>
```
hworld: I don't think you need the ending /, did it add it automatically?

PR #311, opened 2019-08-12
Remove follow games button from onboarding follows step

YLivay:


src/app/components/forms/onboarding/follows/follows.vue
```diff
@@ -4,40 +4,20 @@
...
 					</translate>
 				</h3>
 
 				<p class="text-muted">
-					<translate>
-						Get connected right away with some of the best games on Game Jolt.
-					</translate>
+					<span v-translate>
+						We have fanart, lets plays, guides,
+						<i>DANK</i>
```
nilllzz: Need corporate approval here.

hworld: "Explore fan-created artwork, videos, game guides and more."

hworld:
Looks good once you update the copy

hworld:
You should remove the backend code too so that we don't keep code around that we're not using.

src/app/components/forms/onboarding/follows/follows.vue
YLivay: aww

PR #313, opened 2019-08-15
Update the fixedTime when the date changes

ttbowen:
A small fix in the AppTimeAgo component where the fixedTime isn't updated when the date changes.

I noticed this issue when I viewed a Game Jolt profile and hovered over the join date. First time it's correct but when you navigate to another profile the date is not updated (It shows the joined date for the previous user). This should fix that issue.


hworld:
Looks great, thanks!

PR #318, opened 2019-09-02
Community Manage

YLivay:


src/app/components/forms/community/details/details.vue
```diff
@@ -0,0 +1,38 @@
...
+					max: 100,
+				}"
+				:disabled="!!formModel.game"
+			/>
+			<app-form-control-errors />
+		</app-form-group>
+
+		<app-form-group name="path" :label="$gettext(`URL Path`)">
+			<app-form-control-prefixed-input
+				prefix="gamejolt.com/c/"
```
nilllzz: Can we do "environment.baseurl" here?

src/app/components/forms/community/details/details.ts
```diff
@@ -0,0 +1,18 @@
...
+
+@Component({})
+export default class FormCommunityDetails extends BaseForm<Community> {
+	modelClass = Community;
+	saveMethod = '$saveDetails' as '$saveDetails';
+
+	get communityUrl() {
+		return (
+			'gamejolt.com/c' +
+			'/<b>' +
```
nilllzz: What's with this weird new line? tslint?

src/app/components/forms/community/details/details.vue
```diff
@@ -0,0 +1,38 @@
...
+				data-vv-delay="500"
+				:rules="{
+					pattern: 'urlPath',
+					max: 50,
+					availability: {
+						url: '/web/dash/communities/check-field-availability/path',
+						initVal: model ? model.path : undefined,
+					},
+				}"
+				:disabled="!!formModel.game"
```
nilllzz: The prefixed input does not seem to have a "disabled" prop. Which also means that currently, this field is not disabled for game communities even though it should be.

nilllzz: Just to be clear, the condition is most likely correct, the prop just doesn't get propagated properly.

nilllzz: Prop propagated properly.

src/app/components/forms/community/details/details.ts
YLivay: yes. the thing we use for spell checking.

src/app/components/forms/community/details/details.vue
YLivay: that made my mind tongue numb.

src/app/components/forms/community/details/details.vue
YLivay: i thought about it, but we dont have a const for it without the url scheme. didn't want to add it just for this and then update all the other consts to be based on it. annoying.

src/app/components/forms/community/details/details.vue
YLivay: Also, good catch. since the input is no longer top level native attributes cant be applied directly. guess we'll have to keep that in mind.

PR #320, opened 2019-09-09
Mention suggestions 2

nilllzz:


src/_common/content/content-editor/controls/mention/autocomplete/controls.ts
```diff
@@ -0,0 +1,223 @@
...
+		this.selectedIndex = 0;
+		// If we already have the search results cached, use those and don't query backend.
+		if (ContentEditorMentionCache.hasResults(query)) {
+			this.users = ContentEditorMentionCache.getResults(query);
+			await this.handleInverted();
+		} else {
+			this.users = [];
+
+			// We set a timeout for 200ms here to not send right away when the user is typing fast.
+			this.remoteSuggestionTimeout = setTimeout(async () => {
```
YLivay: First line of the timeout function save the current instance of this.remoteSuggestionTImeout. We need to check if it changed after awaiting the suggestion request to see if we need to abort.

src/_common/content/content-editor/controls/mention/autocomplete/controls.ts
```diff
@@ -0,0 +1,223 @@
...
+
+		this.selectedIndex = 0;
+		// If we already have the search results cached, use those and don't query backend.
+		if (ContentEditorMentionCache.hasResults(query)) {
+			this.users = ContentEditorMentionCache.getResults(query);
+			await this.handleInverted();
+		} else {
+			this.users = [];
+
+			// We set a timeout for 200ms here to not send right away when the user is typing fast.
```
YLivay: Update comment. its waiting for 500ms atm.

src/_common/content/content-editor/controls/mention/autocomplete/controls.ts
```diff
@@ -0,0 +1,223 @@
...
+
+	onClickInsert(user: User) {
+		this.insertUser(user);
+	}
+
+	insertUser(user: User) {
+		if (this.visible && this.canShow) {
+			// start - end include the @query text, it gets replaced with the insertText call.
+			const start = this.view.state.selection.from - this.query.length - 1;
+			const end = this.view.state.selection.from;
```
YLivay: Same issue pointed here:
https://github.com/gamejolt/gamejolt/pull/320/files#diff-cfbbb22adc54bb5903ea5273c034a36dR88

This will fail if we intend to fix the edge case for editing a mention word in the middle.

src/_common/content/content-editor/controls/mention/autocomplete/controls.ts
```diff
@@ -0,0 +1,223 @@
...
+					const searchUsers = User.populate(payload.users);
+
+					// Add to cache
+					if (!ContentEditorMentionCache.hasResults(query)) {
+						ContentEditorMentionCache.addResults(query, searchUsers);
+					}
+
+					// Only process results if the currently active suggestion is still the one we initiated the search with.
+					if (query === this.query) {
+						this.users = searchUsers;
```
YLivay: Do we need to handle the currently selected index? It might now be out of bound.
e.g. if we had 3 suggestions and we skipped the first two suggestions, then added a character which returned only 1 suggestion.

src/_common/content/content-editor/controls/mention/autocomplete/controls.ts
```diff
@@ -0,0 +1,223 @@
...
+					'/web/search/mention-suggestions?q=' + encodeURIComponent(query),
+					undefined,
+					{ detach: true }
+				);
+
+				if (payload.users) {
+					const searchUsers = User.populate(payload.users);
+
+					// Add to cache
+					if (!ContentEditorMentionCache.hasResults(query)) {
```
YLivay: Why do we care to check before setting? If we've already done the query and we have more updated info, whats the harm?

src/_common/content/content-editor/controls/mention/cache.service.ts
```diff
@@ -0,0 +1,20 @@
+import { User } from '../../../../user/user.model';
+
+export default class ContentEditorMentionCache {
+	private static _results: { [query: string]: User[] } = {};
+
+	public static addResults(query: string, users: User[]) {
```
YLivay: Consider renaming to `setResults`

src/_common/content/content-editor/plugins/input-rules/detect-mention-suggestion.ts
```diff
@@ -0,0 +1,27 @@
...
+export function detectMentionSuggestionRule(editor: AppContentEditor) {
+	return new InputRule(
+		/@([\w_-]+)$/i,
+		(
+			state: EditorState<ContentEditorSchema>,
+			_match: string[],
+			_start: number,
+			_end: number
+		) => {
+			// We don't want to insert lists inside code text.
```
YLivay: Update comment

src/_common/content/content-editor/plugins/input-rules/detect-mention-suggestion.ts
```diff
@@ -0,0 +1,27 @@
...
+
+			if (ContentEditorService.checkCurrentNodeIsCode(state)) {
+				return null;
+			}
+
+			editor.canShowMentionSuggestions = 2;
```
YLivay: Explain why 2.

src/_common/content/content-editor/plugins/update-incrementer-plugin.ts
```diff
@@ -22,6 +22,11 @@ export class UpdateIncrementerPlugin {
 		}
 		// If anything in the editor changes (content or selection), make sure we increment, so we can for example reposition controls.
 		this.appEditor.stateCounter++;
+
+		if (this.appEditor.canShowMentionSuggestions > 0) {
```
YLivay: Explain, or at least point to detectMentionSuggestions input rule for more info.

src/_common/content/content-editor/controls/mention/autocomplete/controls.ts
```diff
@@ -0,0 +1,223 @@
...
+	@Watch('stateCounter')
+	private update() {
+		if (this.canShow && this.view instanceof EditorView) {
+			const state = this.view.state;
+			const node = ContentEditorService.getSelectedNode(this.view.state);
+
+			if (node !== null && node.isText) {
+				// Get all text from the beginning of the doc to the end of the selection.
+				// Then from the end of that slice, get the mention text.
+				const slice = state.doc.slice(0, state.selection.from);
```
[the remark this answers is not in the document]
nilllzz: Correct. That is how discord and slack work, and it's much easier to implement honestly.

src/_common/content/content-editor/controls/mention/autocomplete/controls.ts
```diff
@@ -0,0 +1,223 @@
...
+				const text = ContentEditorService.getFragmentText(slice.content);
+				const matches = /@([\w_-]+)$/i.exec(text);
```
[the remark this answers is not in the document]
nilllzz: The following condition is not to check the length of the result, but the length of the regex match result. The first one is the entire match, the second one the capture. So it has to at least be two items so I can make sure the capture worked.

src/_common/content/content-editor/controls/mention/autocomplete/controls.ts
```diff
@@ -0,0 +1,223 @@
...
+			this.users = ContentEditorMentionCache.getResults(query);
+			await this.handleInverted();
+		} else {
+			this.users = [];
+
+			// We set a timeout for 200ms here to not send right away when the user is typing fast.
+			this.remoteSuggestionTimeout = setTimeout(async () => {
+				this.isLoading = true;
+
+				const payload = await Api.sendRequest(
```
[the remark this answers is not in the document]
nilllzz: Why?

src/_common/content/content-editor/controls/mention/autocomplete/controls.ts
nilllzz: It gets reset earlier in this function.

src/_common/content/content-editor/controls/mention/autocomplete/controls.ts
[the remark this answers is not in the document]
YLivay: yeah, i dont think its really important to support as long as it doesnt cause some unexpected issues like choosing to mention user A when in practice user B gets mentioned kind of thing.

src/_common/content/content-editor/controls/mention/autocomplete/controls.ts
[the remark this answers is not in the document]
YLivay: oh yeah, sorry. didnt notice its not an index check haha.
still, the minimum length should be in the regex and you didnt address the reusing bit. later we might come to this and forget this regex is duplicated elsewhere in the code.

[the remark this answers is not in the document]
nilllzz: I did reuse it.

src/_common/content/content-editor/controls/mention/autocomplete/controls.ts
YLivay: is index 0 the "best match"? if so that's fine.

src/_common/content/content-editor/controls/mention/autocomplete/controls.ts
YLivay: comment pls?

src/_common/content/content-editor/controls/mention/autocomplete/controls.ts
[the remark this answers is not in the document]
nilllzz: Oh, and it should start showing from the first character after the @. Of course mentions cannot be less than 2 characters because that's the minimum for usernames, but I think it still makes sense to show suggestions. When selecting the suggestion, it would still be the correct length.

src/_common/content/content-editor/controls/mention/autocomplete/controls.ts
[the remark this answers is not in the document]
YLivay: if between sending the request and getting the response the user navigated away or searched for a different query string then these results are irrelevant and we don't want to apply the result of a search to a state that no longer exists / is irrelevant. updating the cache is fine, but i wouldnt do any changes the users array or anything else.

src/_common/content/content-editor/controls/mention/autocomplete/controls.ts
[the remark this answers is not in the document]
YLivay: the crap? github actually did not show me the new changes until after i hard refreshed.

src/_common/content/content-editor/controls/mention/autocomplete/controls.ts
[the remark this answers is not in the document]
nilllzz: We don't, that's why this exists:
```
// Only process results if the currently active suggestion is still the one we initiated the search with.
if (query === this.query) {
	this.users = searchUsers;
	await this.handleInverted();
}
```

src/_common/content/content-editor/controls/mention/autocomplete/controls.ts
YLivay: disregard this, github goofed.

src/_common/content/content-editor/controls/mention/autocomplete/controls.ts
nilllzz: Yes. The first result (aka index 0), which is determined by the backend, is supposed to always be the best match, and showing at the top of the list.

src/_common/content/content-editor/controls/mention/autocomplete/controls.ts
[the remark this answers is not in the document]
YLivay: the handle inverted showed outside of that if for me cuz git refused to load the changes lol. yeah its fine now.

src/_common/content/content-editor/controls/mention/autocomplete/controls.ts
nilllzz: This is related to your above comment, I'll resolve this.

src/_common/content/content-editor/controls/mention/autocomplete/controls.ts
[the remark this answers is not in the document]
nilllzz: allgud

PR #322, opened 2019-09-10
Fix content set cache

nilllzz:
Fixes gamejolt/issue-tracker#1318

### DIFF
--- a/src/_common/content/content-set-cache.ts
+++ b/src/_common/content/content-set-cache.ts
@@ -50,20 +50,17 @@ export class ContentSetCacheService {
 		context: ContentContext
 	): ContentSetCache {
 		let containerMap = this._caches.get(container);
-
-		if (containerMap) {
-			if (containerMap.has(context)) {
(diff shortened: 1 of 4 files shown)


src/_common/content/content-set-cache.ts
```diff
@@ -38,3 +38,32 @@ export class ContentSetCache {
...
 	}
 }
+
+export class ContentSetCacheService {
+	private static _caches = new WeakMap<
+		ContentContainerModel,
+		Map<ContentContext, ContentSetCache>
+	>();
+
+	public static getCache(
```
YLivay: This entire thing is kind of unclear to me.
So if the cache doesnt exist on the first call, it creates the map and calls itself a second time. the second time it finds the map and initializes the cache and then calls itself a THIRD time to get the cache in the map?

That sounds a bit too complicated to follow for no good reason. Can we get rid of the recursion?

nilllzz: Improved it :)

PR #323, opened 2019-09-11
Featured communities

nilllzz:


src/app/views/discover/home/_banner/banner.ts
```diff
@@ -35,6 +37,13 @@ export default class AppDiscoverHomeBanner extends Vue {
 		return false;
 	}
 
+	get shouldShowJoin() {
```
YLivay: rename to `shouldShowJoinCommunity`

src/app/views/discover/home/_banner/banner.vue
```diff
@@ -62,6 +62,33 @@
...
+									solid
+									:href="item.custom_url"
+									target="_blank"
+									v-app-track-event="`home:banner:custom-community-${item.community.path}`"
+								>
+									{{ item.custom_text }}
+								</app-button>
+
+								<app-community-join-widget
+									v-if="shouldShowJoin"
```
YLivay: maybe instead of not showing for guests, use v-app-auth-required on this?

nilllzz: Idk, this is following what the game follow button does. Should we do it for both?

PR #325, opened 2019-09-19
Community Stuff

YLivay:
* Create communities
* Remove communities
* Improvements to edit
* Resign from moderating a community
* Force community collaborators to follow their own communities


src/_common/community/join-widget/join-widget.vue
```diff
@@ -5,9 +5,10 @@
 		:block="block"
 		:solid="community.is_member || solid"
 		:badge="badge"
-		:disabled="isProcessing"
+		:disabled="isProcessing || (!canJoin && !community.is_member)"
```
nilllzz: Extract this into a getter.

src/_common/fireside/post/post-model.ts
```diff
@@ -321,6 +322,21 @@ export class FiresidePost extends Model implements ContentContainerModel {
 		return this.$_save(`/web/communities/manage/reject/${c.id}`, 'post');
 	}
 
+	$moveChannel(community: Community, channel: CommunityChannel) {
+		const c = this.getTaggedCommunity(community);
+		if (!c) {
+			throw new Error('Cannot move a post to a channel in a community it is not tagged in');
+		}
+
+		if (community.id !== channel.community_id) {
```
nilllzz: This should not be possible, right? And we are catching this in backend anyway iirc.

src/app/views/discover/home/home.vue
```diff
@@ -6,10 +6,8 @@
 
 		<section class="section">
 			<div class="container-xl">
-				<app-discover-home-communities
-					v-if="featuredCommunities.length > 0"
-					:communities="featuredCommunities"
-				/>
+				<!-- v-if="featuredCommunities.length > 0" -->
```
nilllzz: remove comment.

src/_common/community/add-widget/add-widget.vue
```diff
@@ -0,0 +1,51 @@
...
+	theme-prop('color', 'fg-muted')
+
+	&:not(.-disabled):hover
+			theme-prop('border-color', 'highlight')
+			theme-prop('color', 'highlight')
+
+.-disabled
+	cursor: not-allowed
+
+	// Undoes pressy()
```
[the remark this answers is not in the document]
YLivay: Interesting. In general, I think CSS tries being a "forwards-only" kind of thing, so styles to undo some other styles feel like an anti-pattern.

On the hand repeating the pressy() style in each specific item is meh too...

Fuggit, we have sans-margin, maybe we should have sans-pressy haha

src/_common/popper/popper.vue
```diff
@@ -20,6 +20,7 @@
 		@apply-show="onShow"
 		@apply-hide="onHide"
 		@auto-hide="onAutoHide"
+		@contextmenu.native="onContextMenu"
```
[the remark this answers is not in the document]
YLivay: It does not. We should only use it for useful shortcuts, never something that is only reachable through a context menu.

src/_common/fireside/post/post-model.ts
YLivay: Correct, it shouldn't but since this is a function on the post model it by itself can be misused, so I put this guard.

PR #329, opened 2019-10-05
Sectionize community edit

nilllzz:


src/app/views/communities/view/overview/_nav/edit/edit.ts
```diff
@@ -0,0 +1,24 @@
...
+export default class AppCommunitiesViewOverviewNavEdit extends Vue {
+	@Prop(Community)
+	community!: Community;
+
+	isNavExpanded = false;
+
+	readonly Screen = Screen;
+
+	get activeItem() {
+		return 'Overview';
```
YLivay: Do we not want to show which section is currently being viewed?

src/app/views/communities/view/overview/_nav/edit/edit.vue
```diff
@@ -0,0 +1,93 @@
...
+		<div
+			v-if="!Screen.isMobile || isNavExpanded"
+			:class="{
+				'-mobile-nav-container fill-darker anim-fade-in-up': Screen.isMobile,
+			}"
+		>
+			<ol>
+				<app-community-perms
+					:community="community"
+					required="community-media,community-details"
```
YLivay: is media perm needed? the media is currently thumbnail + header which are editable from the page header where you are.

YLivay: Ah, I see.. is it because the thumbnail is squashed into the details page on mobile? Feels kind of odd putting it in details tho, but there isn't enough meat to justify a design section either. hmm...

src/app/views/communities/view/overview/edit/moderators/moderators.ts
```diff
@@ -0,0 +1,97 @@
...
+	collaborators: Collaborator[] = [];
+	activeCollaborator: Collaborator | null = null;
+	isShowingCollaboratorAdd = false;
+
+	readonly Collaborator = Collaborator;
+
+	get isOwner() {
+		// The owner's collaboration is not returned from backend.
+		return this.collaboration === null;
+	}
```
YLivay: It would be cool to put this in the edit store instead of repeating it everywhere. Mostly because the decision to distinguish between owner and collaborators is lack of collaboration entry while owners DO have a collaboration entry, just with the owner role is too arbitrary...

src/app/views/communities/view/overview/_nav/edit/edit.vue
nilllzz: Yeah, didn't want to do that especially because it's mobile only.
We can reorganize that later when we have themes.

src/app/views/communities/view/overview/_nav/edit/edit.ts
nilllzz: Oops forgot to finish this.

PR #334, opened 2019-10-06
Community collab remove comments

nilllzz:


nilllzz:
> Remove comment through moderate is indistinguishable from normal remove as a community collaborator.
> 
> Basically change the condition in the v-if around the "Remove Comment (Moderate)" option in comment widget to `v-if="canRemove"`

I think you are confusing our moderation tools with the community moderation. The "Remove Comment (Moderate)" is not the community moderator. For community moderation, the entry simply says "Remove Comment" like in any other context.

PR #336, opened 2019-10-07
Community share

nilllzz:


src/app/views/communities/view/overview/overview.ts
```diff
@@ -242,6 +253,16 @@ export default class RouteCommunitiesViewOverview extends BaseRouteComponent {
 		return this.communityStates.getCommunityState(this.community);
 	}
 
+	get shareUrl() {
+		return Environment.baseUrl + this.$router.resolve(this.community.routeLocation).href;
+	}
+
+	get shareContent() {
+		return this.$gettextInterpolate('%{ name } Community - Game Jolt', {
```
YLivay: Don't we want some call to action for these? Join / Check out?

PR #337, opened 2019-10-09
Channel customization

nilllzz:


src/app/components/forms/community/channel/edit/edit.ts
```diff
@@ -0,0 +1,62 @@
...
+
+		if (!result) {
+			return;
+		}
+
+		// It's important we save on the base model!
+		// This way we don't overwrite the form model with the current values from the server.
+		// They may have made changes and just want to clear the image and then save their form.
+		// Doing it in this order allows them to do that.
+		await this.model!.$clearBackground();
```
YLivay: The base model is only used to initialize the form model. Changing it does not change the form model.
On the other side of things, the form model should only get assigned back to the model on form submission.

I understand why we clear the background outside of the normal form flow, but why does the order matter?

src/app/components/community/channel/card/card.ts
```diff
@@ -0,0 +1,37 @@
...
+	directives: {
+		AppTooltip,
+	},
+})
+export default class AppCommunityChannelCard extends Vue {
+	@Prop(MediaItem)
+	backgroundItem?: MediaItem;
+
+	@Prop(String)
+	title!: string;
```
YLivay: Just making sure, the reason you pass in these values as props instead of an actual channel model because this is meant to also display the featured channel, which isn't a real channel resource, yeh?

nilllzz: Yes.

PR #339, opened 2019-10-09
Remove message saying you're posting to a community

YLivay:


nilllzz:
Hm, now that I think about it, you think we should make this dismissable instead of straight up removing it? So people can read it the first time and then close it.

PR #342, opened 2019-10-09
Improve community channel pills

YLivay:
Use pills instead of badges, show channel pills in featured feed


src/_common/community/channel/select/select.ts
```diff
@@ -11,5 +16,7 @@ export default class AppCommunityChannelSelect extends Vue {
 	channels!: CommunityChannel[];
 
 	@Emit('input')
-	emitInput(_value: CommunityChannel) {}
+	emitInput(_value: CommunityChannel) {
+		console.log('emitting input');
```
nilllzz: Remove console.log

PR #345, opened 2019-10-09
Post controls lazy

YLivay:
nothing lazy about them, tho. :thinking:


src/app/components/event-item/controls/fireside-post/fireside-post.vue
```diff
@@ -0,0 +1,125 @@
...
+.-controls
+	display: flex
+	justify-content: flex-end
+	align-items: center
+
+.-user-controls
+	flex-grow: 1
+
+.-stats
+	theme-prop('color', 'fg-muted')
```
nilllzz: Use `var(--theme-fg-muted)` instead.

PR #346, opened 2019-10-09
Report posts

YLivay:


src/_common/report/form/form.vue
```diff
@@ -4,7 +4,11 @@
 			<div
 				class="radio"
 				v-if="
-					type === 'Game' || type === 'Comment' || type === 'Forum_Topic' || type === 'Forum_Post'
```
nilllzz: It might be time to put this in an array somewhere and generate this form instead.

PR #350, opened 2019-10-11
Pinned posts

nilllzz:


src/_common/fireside/post/post-model.ts
```diff
@@ -30,6 +31,47 @@ interface FiresidePostPublishedPlatform {
...
+				route.name === 'dash.games.manage.devlog') &&
+			route.params.id.toString() === post.game.id.toString()
+		);
+	}
+
+	if (isInCommunityPinContext(post, route)) {
+		return true;
+	}
+
+	if (route.name === 'profile.overview' && route.params.username === post.user.username) {
```
YLivay: should also check youre looking at the active tab. i think it'll be weird to offer to pin a draft or scheduled post, even if we make that not show up.

src/_common/fireside/post/post-model.ts
```diff
@@ -30,6 +31,47 @@ interface FiresidePostPublishedPlatform {
...
+	return false;
+}
+
+export function isInPinContext(post: FiresidePost, route: Route) {
+	// For the pin option to show, the user has to look at the correct feed for the post's context.
+	// A game post can only be pinned from the game's overview or game's dash feeds.
+	// A user post can only be pinned to the user from the user's overview
+	// 		OR a user post can be pinned to a community channel from a community channel, NOT featured though.
+
+	if (post.game instanceof Game) {
```
YLivay: this is wrong. it would make it impossible to pin a game post posted not as the game owner to your user feed.

src/_common/fireside/post/post-model.ts
```diff
@@ -30,6 +31,47 @@ interface FiresidePostPublishedPlatform {
...
+			route.params.path === community.path &&
+			route.params.channel === channelTitle
+		) {
+			return true;
+		}
+	}
+	return false;
+}
+
+export function isInPinContext(post: FiresidePost, route: Route) {
```
YLivay: I think the better way to go about it is to return which resource is the pin context based on the route and do no permissions check here. then have another function that checks permissions to pin on a given resource separate from where the pin context is.

src/_common/fireside/post/post-model.ts
```diff
@@ -58,6 +100,7 @@ export class FiresidePost extends Model implements ContentContainerModel {
 	url!: string;
 	view_count?: number;
 	expand_count?: number;
+	is_pinned = false; // Field is only gathered, cannot be directly set.
```
YLivay: is_pinned is always returned, this should be `is_pinned!: boolean`.

src/_common/fireside/post/post-model.ts
```diff
@@ -133,6 +176,11 @@ export class FiresidePost extends Model implements ContentContainerModel {
 			this.platforms_published_to = data.platforms_published_to;
 		}
 
+		// Need to assign manually due to default value.
+		if (data.is_pinned) {
```
YLivay: shouldn't need to do this.

src/_common/fireside/post/post-model.ts
```diff
@@ -345,6 +393,24 @@ export class FiresidePost extends Model implements ContentContainerModel {
 		return this.$_save(`/web/posts/manage/publish/${this.id}`, 'firesidePost');
 	}
 
+	$pin(targetModel: string) {
+		const $payload = this.$_save(
+			`/web/posts/manage/toggle-pin/${this.id}/${targetModel}`,
+			'post'
+		);
+		this.is_pinned = true; // Set this manually, the request does not gather this field.
```
YLivay: you shouldn't have to do this. `$_save` assigns the result back to the model.

src/_common/fireside/post/post-model.ts
```diff
@@ -345,6 +393,24 @@ export class FiresidePost extends Model implements ContentContainerModel {
...
+		return $payload;
+	}
+
+	$unpin(targetModel: string) {
+		const $payload = this.$_save(
+			`/web/posts/manage/toggle-pin/${this.id}/${targetModel}`,
+			'post'
+		);
+		this.is_pinned = false; // Set this manually, the request does not gather this field.
```
YLivay: here too

src/app/components/activity/feed/feed.ts
```diff
@@ -208,6 +209,38 @@ export default class AppActivityFeed extends Vue {
...
+		}
+
+		if (
+			eventItem.type === EventItem.TYPE_POST_ADD &&
+			eventItem.action instanceof FiresidePost
+		) {
+			eventItem.action.is_pinned = true;
+		}
+
+		this.onPostEdited(eventItem);
```
YLivay: Not sure what I think about reemitting a pin as an edit post. Maybe we shouldn't emit anything for now.

src/app/components/event-item/controls/fireside-post/extra/extra.ts
```diff
@@ -138,4 +183,24 @@ export default class AppEventItemControlsFiresidePostExtra extends Vue {
 			this.emitRemove();
 		}
 	}
+
+	getPinTargetModel() {
+		let targetModel = 'User';
+		if (this.post.game instanceof Game) {
+			targetModel = 'Game';
```
YLivay: depends on the route, can't assume 'Game'.

src/_common/fireside/post/post-model.ts
nilllzz: It checks this in the extra menu.

src/_common/fireside/post/post-model.ts
nilllzz: Game posts never show up in your user feed though? We removed that yeah?

src/_common/fireside/post/post-model.ts
nilllzz: There are no permission checks in this though?

src/_common/fireside/post/post-model.ts
nilllzz: In the backend endpoint it's pretty difficult to figure out the context, that's why this is.

src/_common/fireside/post/post-model.ts
nilllzz: In the backend endpoint it's pretty difficult to figure out the context, that's why this is.

src/_common/fireside/post/post-model.ts
nilllzz: I did this for backwards comp.

src/_common/fireside/post/post-model.ts
nilllzz: Fixed.

src/app/components/event-item/controls/fireside-post/extra/extra.ts
nilllzz: Fixed.

src/_common/fireside/post/post-model.ts
```diff
@@ -358,7 +358,22 @@ export class FiresidePost extends Model implements ContentContainerModel {
 			}
 		}
 
-		return this.$_save(`/web/posts/manage/save/${this.id}`, 'firesidePost', options);
+		return new Promise(async resolve => {
```
YLivay: Why wrap this in a promise?

PR #360, opened 2019-10-15
add error message for private yt videos

nilllzz:


src/app/components/forms/post/post.ts
```diff
@@ -468,6 +470,17 @@ export default class FormPost extends BaseForm<FormPostModel>
 		Object.assign(this.model, this.formModel);
 	}
 
+	onSubmitError($payload: any) {
+		if ($payload.errors.youtube_private) {
```
YLivay: Change to video_unavailable, we can't be sure for what reason it's unavailable. could be a removed video, region locked, private whatever.

PR #365, opened 2019-10-21
Add support to move posts when removing a channel

YLivay:


src/app/components/community/remove-channel/remove-channel.vue
```diff
@@ -0,0 +1,74 @@
...
+					<app-jolticon icon="remove" />
+				</a>
+			</div>
+		</app-expand>
+
+		<p class="help-block">
+			<translate>Removing a channel is irreversible. Once it's gone, it's gone for good.</translate>
+			<template v-if="moving">
+				<br />
+				<translate>It might take a few minutes for the posts to show in the new channel.</translate>
```
nilllzz: I would change this to "a few moments" instead.

src/app/components/community/remove-channel/remove-channel.ts
```diff
@@ -0,0 +1,69 @@
...
+			await this.channel.$remove(moveToChannel);
+			success = true;
+		} catch (e) {
+			Growls.error(
+				this.$gettext('Could not remove channel for some reason. Try again later!')
+			);
+		}
+
+		if (success) {
+			this.emitRemoved(moveToChannel);
```
nilllzz: Just move this at the end of the `try` and the remove `success`.

src/app/views/communities/view/overview/edit/channels/channels.ts
```diff
@@ -56,26 +56,11 @@ export default class RouteCommunitiesViewEditChannels extends BaseRouteComponent
...
-			return;
-		}
-
-		try {
-			await channel.$remove();
-		} catch (e) {
-			console.error(e);
-			Growls.error('Could not remove channel');
-		}
+		await CommunityRemoveChannelModal.show(this.community, channel);
```
nilllzz: I think we should redirect to the new channel when moving posts to the a channel.

src/app/components/community/remove-channel/remove-channel.ts
YLivay: I don't want errors that may happen during emit to be caught by this catch, and since you're not a fan of multiple return statements for flow control, i figured this just works.

src/app/views/communities/view/overview/edit/channels/channels.ts
YLivay: I think we shouldn't, especially since posts will not show immediately, and they were in the middle of management, and not content creation.

PR #367, opened 2019-10-22
Fix community themes glitching out after editing

YLivay:


src/app/components/forms/community/community.ts
```diff
@@ -27,7 +31,15 @@ export default class FormCommunity extends BaseForm<Community> implements FormOn
 	@ThemeState
 	userTheme!: ThemeStore['userTheme'];
 
+	onInit() {
```
nilllzz: This is unrelated, and fixes that it doesn't show the browser warning when navigating away, right?

PR #369, opened 2019-10-23
Suppress feature notification

nilllzz:


src/app/components/grid/client.service.ts
```diff
@@ -81,6 +84,13 @@ export class GridClient {
 	bootstrapReceived = false;
 	bootstrapTimestamp = 0;
 
+	/**
+	 * Store ids of posts the user has featured.
+	 * The Grid client will ignore any incoming feature notifications for posts recorded here,
+	 * because users that feature posts should not get notified about those exact posts.
+	 */
+	featuredPostIds: number[] = [];
```
YLivay: Use a `Set` for this.

YLivay: also, can't this be `private`?

nilllzz: A set? You mean a setter?
Can you tell me where we would need to access a setter for this variable?

YLivay: a capital S `Set`. as in `Set<number>`

nilllzz: Ah sure, why not.

PR #370, opened 2019-10-23
Descriptive form labels

nilllzz:


YLivay:
next time these types of changes can go directly on master since we talked about them already.

PR #372, opened 2019-10-23
Improve component for community moderators

YLivay:


src/app/components/community/sidebar/sidebar.ts
```diff
@@ -49,18 +49,37 @@ export default class AppCommunitySidebar extends Vue {
...
 	app!: Store['app'];
 
-	collaboratorListCollapsed = false;
+	currentCollaborators: User[] = [];
+	currentCollaboratorCount = 0;
+
+	collaboratorListCollapsed = true;
 	isLoadingMoreCollaborators = false;
 	isShowingShare = false;
+	loadedAll = false;
```
nilllzz: I don't understand what this does. It doesn't get set to `true` ever.

src/app/components/community/sidebar/sidebar.ts
```diff
@@ -49,18 +49,37 @@ export default class AppCommunitySidebar extends Vue {
...
+
+	readonly GJ_IS_CLIENT = GJ_IS_CLIENT;
+
+	@Emit('collaborators')
+	emitCollaborators(_collaborators: User[]) {}
```
nilllzz: The emit doesn't seem to be used by anything?

src/app/components/community/sidebar/sidebar.ts
YLivay: gud catch, should've set to true when loading

YLivay: Also renamed to be clearer what this is loading since sidebar does multiple things

src/app/components/community/sidebar/sidebar.ts
YLivay: Yeah, we'll reimplement once needed.

PR #387, opened 2019-11-01
Tweak the profile trophy designs and stuff.

hworld:


src/app/views/profile/trophies/_nav/nav.vue
```diff
@@ -37,45 +37,19 @@
...
+				<template v-slot="{ item }">
+					<translate v-if="!item">Choose a game...</translate>
+					<template v-else>
+						<span class="badge" :class="{ 'badge-notice': gameHasUnviewedTrophies(item.id) }">
+							{{ item.trophyCount | number }}
 						</span>
-					</router-link>
-				</li>
-			</ul>
+						{{ item.title }}
```
nilllzz: This needs to be shortened if no space, otherwise it'll look like this:  
image

src/app/views/profile/trophies/game/game.vue
```diff
@@ -1,55 +1,36 @@
...
-					<router-link
+		<div class="row" v-if="game">
+			<div v-if="completion" class="col-sm-7">
+				<app-trophy-completion
+					class="-completion"
+					:total="completion.totalCount"
+					:achieved="completion.achievedCount"
+					:experience="completion.experience"
+				/>
+				<p>
```
nilllzz: I think these buttons should show even if you are looking at someone else's profile.

src/app/components/score/scoreboard-selector/scoreboard-selector.vue
```diff
@@ -1,62 +1,14 @@
...
-				<h5 class="list-group-item-heading sans-margin-bottom">
-					<strong>{{ currentTable.name }}</strong>
-				</h5>
-				<p class="list-group-item-text">
-					{{ currentTable.description }}
-				</p>
-			</div>
-		</div>
-	</div>
+	<app-list-group-selector :items="tables" :current="currentTable" @change="selectTable($event)">
```
YLivay: Doesn't seem to scroll with the content.
image

src/_common/list-group/selector/selector.vue
```diff
@@ -0,0 +1,32 @@
+<template>
+	<div class="list-group">
```
YLivay: Doesn't center correctly with long content
image

hworld:
> image
> On mobile, there isn't enough margin between the items and the buttons.

I didn't think this was too big of a deal honestly. I actually don't like the icon too much, since it looks like an error to me, and would actually instead get rid of the icon and just keep the pink border as the indication that it's new. Thoughts on that solution?

> image
> On desktop it seems like there is too much padding between the hr and the thumbnail.

I thought that my code removed this margin. Can you check, maybe it needs an `!important` or something... It's the stylus code within the src/app/views/profile/trophies/game/game.vue file.

src/app/views/profile/trophies/_nav/nav.vue
hworld: This is kind of an interesting issue. We probably need to fix in the list-group styling, since it's all within that component. Maybe try something with overflow-wrap? https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap

src/app/views/profile/trophies/game/game.vue
hworld: Yeah, I agree, this is a bug then

src/app/components/score/scoreboard-selector/scoreboard-selector.vue
hworld: Current issue, not something this PR introduced.

src/_common/list-group/selector/selector.vue
hworld: Current issue, not something this PR introduced.

src/app/views/profile/trophies/_nav/nav.vue
nilllzz: I'm cool with this for now, people usually put a space in their name somewhere haha

src/app/views/profile/trophies/overview/overview.ts
```diff
@@ -70,41 +76,127 @@ export default class RouteProfileTrophiesOverview extends BaseRouteComponent {
...
+		const siteTrophies = allTrophies.filter(i => i instanceof UserSiteTrophy);
+
+		this.canLoadMore =
+			(loadedGameTrophies.length > 0 &&
+				gameTrophies.length > 0 &&
+				gameTrophies.length % PAGE_SIZE === 0) ||
+			(loadedSiteTrophies.length > 0 &&
+				siteTrophies.length > 0 &&
+				siteTrophies.length % PAGE_SIZE === 0);
 	}
```
YLivay: Add comments

src/app/views/profile/trophies/overview/overview.ts
```diff
@@ -20,6 +21,8 @@ type TrophyEntry = {
 	trophies: UserBaseTrophy[];
 };
 
+const PAGE_SIZE = 12;
```
YLivay: rename to `PAGE_SIZE_PER_TROPHY_TYPE` and add a comment saying 12 is chosen arbitrarily (and not chosen to satisfy any design layout constraints).

YLivay: actually, this should be returned from the backend, otherwise you might not allow loading more when you should be able to. could tweak without having to push frontend.

PR #392, opened 2019-11-06
Reactive posts

nilllzz:


src/app/components/post/view/view.ts
```diff
@@ -81,4 +96,68 @@ export default class AppPostView extends Vue {
...
+				this.startedListeningGrid = true;
+				this.grid.startListenPost(this.post, this.handleGridEvent);
+			}
+		}, 2000);
+	}
+
+	handleGridEvent(event: GridEvent, payload: GridEventPayload) {
+		// If the logged in user has taken an action, don't update:
+		// That change is already visible (like++, poll vote etc)
+		if (this.app.user && this.app.user.id === payload.user_id) {
```
YLivay: Note this breaks if the user has multiple tabs on the same resource.

src/app/components/post/view/view.ts
```diff
@@ -81,4 +96,68 @@ export default class AppPostView extends Vue {
...
+				this.post.like_count++;
+				break;
+
+			case 'unlike':
+				this.post.like_count--;
+				break;
+		}
+	}
+
+	destroyed() {
```
YLivay: Check if it actually connected to the post. Also, did you check navigating between different posts?

src/app/components/post/view/view.ts
```diff
@@ -81,4 +96,68 @@ export default class AppPostView extends Vue {
...
+					this.startListenInterval = null;
+				}
+				return;
+			}
+
+			if (this.grid && this.grid.connected) {
+				this.startedListeningGrid = true;
+				this.grid.startListenPost(this.post, this.handleGridEvent);
+			}
+		}, 2000);
```
YLivay: This is ridiculous, i think we need to be able to just queue and abort grid operations through the grid service instead of forcing anything that requires a grid subscription to set this death contraption up.

PR #394, opened 2019-11-24
Ad control

YLivay:


YLivay:
I didn't go git spelunking. iirc it was used for the previous google ads one. we use a different component now for playwire.

PR #407, opened 2020-01-19
Move community about button

nilllzz:


src/app/views/communities/view/overview/edit/blocks/blocks.vue
```diff
@@ -1,7 +1,7 @@
 <template>
 	<div>
 		<h2 class="section-header">
-			<translate>Blocks</translate>
+			<translate>Blocked Users</translate>
```
hworld: Woohoo!

src/app/views/communities/view/overview/overview.ts
```diff
@@ -107,13 +106,8 @@ export default class RouteCommunitiesViewOverview extends BaseRouteComponent {
...
 
 	feed: ActivityFeedView | null = null;
-	knownMembers: User[] = [];
-	knownMemberCount = 0;
 	finishedLoading = false;
-	owner: User | null = null;
-	collaborators: User[] = [];
-	collaboratorCount = 0;
-	initialCollaboratorCount = 0;
+	sidebarData: CommunitySidebarData | null = null;
```
hworld: This can be passed as a prop from the parent "view" component. This way we only need to fetch once at the top-level. The way you do it is define a prop here, and then pass the prop into the <router-view> tag in the "view" component file. Vue-router will actually pass that prop data through into this component.

src/app/components/community/sidebar/sidebar-data.ts
```diff
@@ -0,0 +1,32 @@
+import { Community } from '../../../../_common/community/community.model';
+import { User } from '../../../../_common/user/user.model';
+
+export class CommunitySidebarData {
```
hworld: Why the need for all the `!` in the properties? When is payload ever undefined? It feels really messy to have the type annotations being potentially wrong...

YLivay: Gud call

src/app/components/community/sidebar/sidebar-data.ts
```diff
@@ -0,0 +1,32 @@
...
+		if ($payload) {
+			this.knownMembers = User.populate($payload.knownMembers || []);
+			this.knownMemberCount = $payload.knownMemberCount || 0;
+
+			if ($payload.owner) {
+				this.owner = new User($payload.owner);
+			}
+			if ($payload.collaborators) {
+				this.collaborators = User.populate($payload.collaborators);
+			}
```
YLivay: the owner and collaborators are always returned (even if collaborators are an empty array) so I think we should be throwing if these conditions are not satisfied.

src/app/components/community/sidebar/sidebar.ts
```diff
@@ -79,35 +61,37 @@ export default class AppCommunitySidebar extends Vue {
...
 	}
 
 	get membersYouKnowCount() {
-		return number(this.knownMemberCount);
+		return number(this.data.knownMemberCount);
 	}
 
 	get shareUrl() {
-		return Environment.baseUrl + this.$router.resolve(this.community.routeLocation).href;
+		return Environment.baseUrl + this.$router.resolve(this.data.community.routeLocation).href;
```
YLivay: Check reactivity. now that we're passing a community through a data object from a payload and not directly from a prop we have two separate instances of the community to react on.

Maybe we should keep taking the community as a prop since the community is the "subject" the sidebar is attached to?

src/app/components/community/sidebar/sidebar.ts
```diff
@@ -79,35 +61,37 @@ export default class AppCommunitySidebar extends Vue {
...
-		if (this.owner) {
-			mods.push(this.owner);
+		if (this.data.owner) {
+			mods.push(this.data.owner);
 		}
 		if (this.currentCollaborators) {
 			if (this.collaboratorListCollapsed) {
-				mods.push(...this.currentCollaborators.slice(0, this.initialCollaboratorCount));
+				mods.push(
+					...this.currentCollaborators.slice(0, this.data.initialCollaboratorCount)
```
YLivay: Again, check reactivity. I think if you accepted a collaboration you wouldn't show in the list until after a refresh with this change.

src/app/views/communities/view/overview/overview.ts
nilllzz: We will still fetch it, just not process it due to backwards comp, right?
Old clients will still need the data to be returned from the overview endpoint.

nilllzz: But sure, I'll pass it down through the router view.

src/app/components/community/sidebar/sidebar.ts
nilllzz: Pretty sure we never supported this?
I just tested this in master and it didn't seem to update the list when you click accept.

nilllzz: i did add this though, and it works without changing the bit of code highlighted here.
I think Vue reactivity goes into nested data like this unless the entire instance is swapped out.
Since it's passed in as a prop, Vue handles it correctly.

src/app/components/community/sidebar/sidebar.ts
```diff
@@ -68,22 +54,22 @@ export default class AppCommunitySidebar extends Vue {
 
 	readonly GJ_IS_CLIENT = GJ_IS_CLIENT;
 
-	@Watch('collaborators', { immediate: true })
+	@Watch('data.collaborators', { immediate: true })
```
hworld: Shouldn't these two watches just be getters?

nilllzz: I don't think so, it's because this component holds its own state for those two properties, but wants to overwrite with incoming prop data when it is provided with it.

PR #414, opened 2020-02-10
Login cover credits

hapurubokka:


src/auth/components/game-img/game-img.ts
```diff
@@ -0,0 +1,37 @@
+import Vue from 'vue';
+import Component from 'vue-class-component';
+import { State } from 'vuex-class';
+import AppGameThumbnailImg from '../../../_common/game/thumbnail-img/thumbnail-img.vue';
+import { Store } from '../../store/index';
+
+@Component({
+	name: 'GameCoverCredits',
```
hworld: No need for name in the components. Only the route components.

src/auth/components/game-img/game-img.ts
```diff
@@ -0,0 +1,37 @@
...
+	components: {
+		AppGameThumbnailImg,
+	},
+})
+export default class GameCoverCredits extends Vue {
```
hworld: Always name component classes with App as a prefix. Also, make sure the filename matches the class name. So the filename should be `game-cover-credits` and the class should be `AppGameCoverCredits`

src/auth/components/game-img/game-img.vue
```diff
@@ -0,0 +1,108 @@
+import { ThemeStore } from '../../../_common/theme/theme.store';
+<template>
+	<div id="cover-img-credits">
```
hworld: don't use ID. always use classes. the standard I use for these is that the top-level div would be the name of the component in kebab-case, so `class="game-cover-credits`.

src/auth/components/game-img/game-img.vue
```diff
@@ -0,0 +1,108 @@
...
+			>
+			<span class="text-muted -follows">{{ gameFollows }} followers</span>
+		</div>
+		<a class="-img" :href="gameUrl">
+			<app-game-thumbnail-img :game="coverGame" />
+		</a>
+	</div>
+</template>
+
+<style scoped>
```
hworld: always use stylus as the lang for style code in GJ
With stylus, don't add the {} brackets, and don't add `;` at the end of lines

src/auth/components/game-img/game-img.vue
```diff
@@ -0,0 +1,108 @@
...
+#cover-img-credits {
+	position: fixed;
+	right: 17px;
+	bottom: 0;
+	padding: 5px;
+	margin: 20px;
+
```
hworld: get rid of any extra lines, no groupings of properties

src/auth/components/game-img/game-img.vue
```diff
@@ -0,0 +1,108 @@
...
+	display: flex;
+
+	background: #262b2b;
+	border-radius: 8px;
+}
+
+#cover-img-credits a.-title:hover {
```
hworld: since it's stylus, you would put these rules indented under the top-level class, however, since you named it .-title, just use that as the selector: `.-title:hover`

src/auth/components/game-img/game-img.vue
```diff
@@ -0,0 +1,108 @@
...
+	text-decoration: underline;
+}
+
+#cover-img-credits {
```
hworld: This should go up above

src/auth/components/game-img/game-img.vue
```diff
@@ -0,0 +1,108 @@
...
+	border-radius: 8px;
+}
+
+.-img:hover {
+	/* box-shadow: lightgray 0px 0px 2px; */
+
+	box-shadow: 0px 0px 3px var(--theme-bi-bg) !important;
+}
+
+.flow-col {
```
hworld: prefix with `-`

src/auth/components/game-img/game-img.vue
```diff
@@ -0,0 +1,108 @@
...
+.-dev {
+	font-size: 13px;
+	margin: 3px;
+}
+
+.-follows {
+	font-size: 13px;
+}
+
+/* .mask {
```
hworld: remove

src/auth/components/game-img/game-img.vue
```diff
@@ -0,0 +1,108 @@
...
+}
+
+.-img {
+	width: 145px;
+	margin-left: 10px;
+	transition: 0.25s;
+	border-radius: 8px;
+}
+
+.-img:hover {
```
hworld: always remove commented code

src/auth/components/game-img/game-img.ts
```diff
@@ -0,0 +1,37 @@
...
+	@State coverGame: Store['coverGame'];
+
+	get gameTitle() {
```
hworld: just include these in the template within the interpolations directly, no need for getters here since they're simple access

src/auth/components/game-img/game-img.vue
hworld: put a `v-if` on this div to make sure `coverGame` exists: `v-if="coverGame"`

src/auth/components/game-img/game-img.vue
```diff
@@ -0,0 +1,108 @@
+import { ThemeStore } from '../../../_common/theme/theme.store';
```
hworld: This accidentally was included here, you should remove

src/auth/components/game-img/game-img.vue
```diff
@@ -0,0 +1,108 @@
...
+<template>
+	<div id="cover-img-credits">
+		<div class="flow-col">
+			<a class="link-unstyled -title" :href="gameUrl">{{ gameTitle }}</a>
```
hworld: always use router-link, you would just change `href` to `to`: `<router-link class="-title link-unstyled" :to="gameUrl">...`

src/auth/components/game-img/game-img.vue
```diff
@@ -0,0 +1,108 @@
...
+			<a class="link-muted -dev" :href="devUrl"
```
hworld: ditto to above

src/auth/components/game-img/game-img.vue
```diff
@@ -0,0 +1,108 @@
...
+				>by <strong>{{ developerName }}</strong></a
```
hworld: maybe put a new line before the by and after the strong? prettier might format it better

src/auth/components/game-img/game-img.vue
```diff
@@ -0,0 +1,108 @@
...
+			>
+			<span class="text-muted -follows">{{ gameFollows }} followers</span>
```
hworld: put the scoped class first, like `-followed text-muted`

src/auth/components/game-img/game-img.vue
```diff
@@ -0,0 +1,108 @@
...
+		</div>
+		<a class="-img" :href="gameUrl">
```
hworld: always router link

src/auth/components/game-img/game-img.vue
```diff
@@ -0,0 +1,108 @@
...
+			<app-game-thumbnail-img :game="coverGame" />
+		</a>
+	</div>
+</template>
+
+<style scoped>
+#cover-img-credits {
```
hworld: The parent should position this. This component should just render itself, this way it can put included anywhere

src/auth/components/game-img/game-img.vue
```diff
@@ -0,0 +1,108 @@
...
+
+#cover-img-credits a.-title:hover {
+	text-decoration: underline;
+}
+
+#cover-img-credits {
+	text-align: right;
+}
+
+#cover-img-credits a:hover {
```
hworld: no need for the ID selector here, just a:hover. vue will scope it properly to this component

src/auth/views/auth/auth.vue
```diff
@@ -48,6 +48,7 @@
 					<router-view />
 				</div>
 			</div>
+			<game-img-credits :game="coverGame" />
```
hworld: put a new line above, because I'm picky

PR #415, opened 2020-02-11
User blocks

nilllzz:


src/app/components/forms/user/block/block.ts
```diff
@@ -1,26 +1,42 @@
...
+	components: {
+		AppFormControlToggle,
+	},
 })
 export default class FormUserBlock extends BaseForm<BlockData> implements FormOnSubmit {
 	resetOnSubmit = true;
 
 	async onSubmit() {
 		const response = await Api.sendRequest(`/web/dash/blocks/add`, this.formModel);
 		if (response.success) {
```
hworld: Hey, I just noticed this. I think you're supposed to do just the API call in onSubmit, and then you should make a new onSubmitSuccess method for handling the success and you can push all this into that.

src/_common/block/form/form.ts
```diff
@@ -1,17 +1,27 @@
 import Component from 'vue-class-component';
 import { Prop } from 'vue-property-decorator';
 import { Api } from '../../api/api.service';
+import AppFormControlToggle from '../../form-vue/control/toggle/toggle.vue';
 import { BaseForm, FormOnSubmit } from '../../form-vue/form.service';
 import { User } from '../../user/user.model';
 
-@Component({})
-export default class AppBlockForm extends BaseForm<any> implements FormOnSubmit {
+type BlockData = {
```
hworld: make an interface and call it FormModel, I think that's our convention, yeah?

hworld:
Nothing really that important to "fix." I added a couple more comments, then feel free to just merge this.

src/_common/block/form/form.ts
nilllzz: We do both interface and type, I'll rename it tho.

PR #434, opened 2020-03-04
Update game media modal

nilllzz:
adds live previews while editing urls
accepts 3 types of input for sketchfab (new urls, old urls, model ids)
adjusted text to be equal between post and game modals
added old url support for post modal

### DIFF
--- a/src/app/components/forms/game/sketchfab/sketchfab.ts
+++ b/src/app/components/forms/game/sketchfab/sketchfab.ts
@@ -1,17 +1,56 @@
-import { BaseForm, FormOnInit } from '../../../../../_common/form-vue/form.service';
+import { Component, Prop } from 'vue-property-decorator';
+import { BaseForm, FormOnInit, FormOnSubmit } from '../../../../../_common/form-vue/form.service';
 import { Game } from '../../../../../_common/game/game.model';
 import { GameSketchfab } from '../../../../../_common/game/sketchfab/sketchfab.model';
-import { Component, Prop } from 'vue-property-decorator';
+import AppSketchfabEmbed from '../../../../../_common/sketchfab/embed/embed.vue';
 
-@Component({})
-export default class FormGameSketchfab extends BaseForm<GameSketchfab> implements FormOnInit {
+@Component({
+	components: {
+		AppSketchfabEmbed,
+	},
+})
+export default class FormGameSketchfab extends BaseForm<GameSketchfab>
+	implements FormOnInit, FormOnSubmit {
 	@Prop(Game) game!: Game;
+
+	readonly SKETCHFAB_URL_REGEX_1 = /^(?:(?:https:\/\/)?(?:www.)?(?:sketchfab.com\/3d-models\/(?:[a-z0-9]+-)+([0-9a-f]{32})\/?))$/i;
+	readonly SKETCHFAB_URL_REGEX_2 = /^(?:(?:https:\/\/)?(?:www.)?(?:sketchfab.com\/models\/([0-9a-f]{32})\/?))$/i;
+	readonly SKETCHFAB_FIELD_REGEX = /^(?:(?:https:\/\/)?(?:www.)?(?:sketchfab.com\/3d-models\/(?:[a-z0-9]+-)+([0-9a-f]{32})\/?))$|^(?:(?:https:\/\/)?(?:www.)?(?:sketchfab.com\/models\/([0-9a-f]{32})\/?))$|^([0-9a-f]{32})$/i;
 
 	modelClass = GameSketchfab;
 	resetOnSubmit = true;
 	warnOnDiscard = false;
 
+	get sketchfabId() {
+		for (const regex of [this.SKETCHFAB_URL_REGEX_1, this.SKETCHFAB_URL_REGEX_2]) {
+			const urlMatches = regex.exec(this.formModel.sketchfab_id.trim());
+			if (urlMatches !== null && urlMatches.length === 2) {
+				const id = urlMatches[1];
+				return id;
+			}
+		}
+		return this.formModel.sketchfab_id;
+	}
+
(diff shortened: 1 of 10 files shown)


src/app/components/forms/game/video/video.ts
```diff
@@ -36,17 +58,9 @@ export default class FormGameVideo extends BaseForm<FormModel> implements FormOn
...
-			if (youtubeMatch) {
-				this.setField('type', GameVideo.TYPE_YOUTUBE);
-				this.setField('url', youtubeMatch[4]);
-			} else if (vimeoMatch) {
-				this.setField('type', GameVideo.TYPE_VIMEO);
-				this.setField('url', vimeoMatch[4]);
-			}
+		if (url && this.hasValidVideoUrl) {
+			this.setField('type', this.videoData!.type);
+			this.setField('url', this.videoData!.id);
```
YLivay: getters are only cached for vue rendering. the videoData would still be called 3 times in this. lets just get the videoData once and work off of it, yeah?

src/app/components/forms/game/video/video.ts
```diff
@@ -19,6 +22,25 @@ export default class FormGameVideo extends BaseForm<FormModel> implements FormOn
...
 	GameVideo = GameVideo;
 
+	get hasValidVideoUrl() {
+		return !!this.videoData;
+	}
+
+	get videoData() {
+		const url = this.formModel._url;
+		if (!url) {
+			return undefined;
```
YLivay: use null, don't use undefined as a concrete return value.

src/app/components/forms/game/sketchfab/sketchfab.ts
```diff
@@ -1,17 +1,56 @@
...
+@Component({
+	components: {
+		AppSketchfabEmbed,
+	},
+})
+export default class FormGameSketchfab extends BaseForm<GameSketchfab>
+	implements FormOnInit, FormOnSubmit {
 	@Prop(Game) game!: Game;
 
+	readonly SKETCHFAB_URL_REGEX_1 = /^(?:(?:https:\/\/)?(?:www.)?(?:sketchfab.com\/3d-models\/(?:[a-z0-9]+-)+([0-9a-f]{32})\/?))$/i;
```
YLivay: Are you sure the first part after the `3d-models` bit isn't part of the hash that would identify the model?
e.g. `artist-abcde124` instead of just `abcde124`

YLivay: also the `.` in `www.` should be escaped to be `www\.` Same with the `.` in `sketchfab.com`

src/app/components/forms/game/sketchfab/sketchfab.ts
```diff
@@ -1,17 +1,56 @@
...
+	readonly SKETCHFAB_URL_REGEX_2 = /^(?:(?:https:\/\/)?(?:www.)?(?:sketchfab.com\/models\/([0-9a-f]{32})\/?))$/i;
+	readonly SKETCHFAB_FIELD_REGEX = /^(?:(?:https:\/\/)?(?:www.)?(?:sketchfab.com\/3d-models\/(?:[a-z0-9]+-)+([0-9a-f]{32})\/?))$|^(?:(?:https:\/\/)?(?:www.)?(?:sketchfab.com\/models\/([0-9a-f]{32})\/?))$|^([0-9a-f]{32})$/i;
```
YLivay: This is kind of unreadble now. Also, whats the difference between url regex and field regex? can you put a little comment explaining the use case for these?

src/app/components/forms/post/post.ts
```diff
@@ -117,8 +117,9 @@ export default class FormPost extends BaseForm<FormPostModel>
 	};
 
 	readonly YOUTUBE_URL_REGEX = /^(?:https?:\/\/)?(?:www\.)?(?:youtube\.com\/watch\?v=|youtu\.be\/)([a-zA-Z0-9_\-]{11})(?:&.+)*$/i;
-	readonly SKETCHFAB_URL_REGEX = /^(?:(?:https:\/\/)?(?:www.)?(?:sketchfab.com\/3d-models\/(?:[a-z0-9]+-)+([0-9a-f]{32})\/?))$/i;
-	readonly SKETCHFAB_FIELD_REGEX = /^(?:(?:https:\/\/)?(?:www.)?(?:sketchfab.com\/3d-models\/(?:[a-z0-9]+-)+([0-9a-f]{32})\/?))$|^([0-9a-f]{32})$/i;
+	readonly SKETCHFAB_URL_REGEX_1 = /^(?:(?:https:\/\/)?(?:www.)?(?:sketchfab.com\/3d-models\/(?:[a-z0-9]+-)+([0-9a-f]{32})\/?))$/i;
+	readonly SKETCHFAB_URL_REGEX_2 = /^(?:(?:https:\/\/)?(?:www.)?(?:sketchfab.com\/models\/([0-9a-f]{32})\/?))$/i;
+	readonly SKETCHFAB_FIELD_REGEX = /^(?:(?:https:\/\/)?(?:www.)?(?:sketchfab.com\/3d-models\/(?:[a-z0-9]+-)+([0-9a-f]{32})\/?))$|^(?:(?:https:\/\/)?(?:www.)?(?:sketchfab.com\/models\/([0-9a-f]{32})\/?))$|^([0-9a-f]{32})$/i;
```
YLivay: I don't think theres a justification to duplicate this fresh hell. We should probably refactor this to some sketchfab helper/util.

src/app/components/forms/game/sketchfab/sketchfab.ts
nilllzz: The new Sketchfab urls look like this:
https://sketchfab.com/3d-models/lowpoly-80s-tv-70001ecdc29c43ceba1d766f09fe79c0
which have the name slugified and then the id of the model.

PR #436, opened 2020-03-09
Change user follow modal to an expanding div in post controls

hapurubokka:
- Changed the way that we ask if people want to follow the user of a post that was liked. 
- Displays in an expanding div instead of a modal.
- Will now display on mobile devices as well.

**comment-store.ts**
fixed console errors if 'hydrated.data' was null

**view.ts**
created state for if the we should be showing a user-follow component (AppEventItemControlsUserFollow)

**view.vue**
'app-event-item-controls' is now being passed 'show-follow' on the post view, so we will show on posts now.

**widget.ts**
removed all logic about showing follow modals. Instead, the like widget will now only trigger an 'emitChange(boolean)' event when you like/unlike something.

**fireside-post, ts & vue**
instead of passing whether we should show a follow modal to 'widget.ts', we are now only chaining a change on when a post is liked/unliked.

**event-item.vue**
we are now passing 'feed' and 'item' to 'controls.ts' so it can check and set the state on if we should be showing 'user-follow' or not.

**controls, ts & vue**
added 'user-follow' component to display when a post is liked. This component is now what is responsible for deciding if we should be showing the 'user-follow' expansion.

**user-follow**
this is replacing the modal for following users when a post is liked. Now it will expand in whatever component it is being added to, rather than creating a seperate modal to display.


src/_common/fireside/post/like/widget/widget.ts
```diff
@@ -1,15 +1,13 @@
 import Vue from 'vue';
-import { Component, Prop } from 'vue-property-decorator';
+import { Component, Emit, Prop } from 'vue-property-decorator';
```
hworld: Get rid of the AppPopper component that's included in this component.

src/app/components/event-item/controls/fireside-post/fireside-post.vue
```diff
@@ -9,7 +9,11 @@
 
 		<div class="-controls">
 			<div class="-user-controls" v-if="showUserControls">
-				<app-fireside-post-like-widget :post="post" :show-user-follow="showUserFollow" trans />
+				<app-fireside-post-like-widget
+					:post="post"
+					@change="emitLikeChange"
```
hworld: put events last, after the trans prop

PR #437, opened 2020-03-10
Community channel permissions

nilllzz:


src/_common/community/channel/channel.model.ts
```diff
@@ -9,12 +10,17 @@ export class CommunityChannel extends Model {
 	sort!: number;
 	background?: MediaItem;
 
+	permissions!: ChannelPermissions;
+	_permissions!: string;
```
YLivay: I don't like this. lets talk about it later?

src/_common/community/channel/select/select.ts
```diff
@@ -17,4 +17,8 @@ export default class AppCommunityChannelSelect extends Vue {
 
 	@Emit('input')
 	emitInput(_value: CommunityChannel) {}
+
+	get validChannels() {
+		return this.channels.filter(i => i.permissions.canPerform('posts'));
```
YLivay: Add action consts on channel permissions or channel model.

src/app/components/forms/community/channel/edit/edit.ts
```diff
@@ -2,16 +2,12 @@ import { Component, Prop } from 'vue-property-decorator';
...
-	BaseForm,
-	FormOnLoad,
-	FormOnSubmitSuccess,
-} from '../../../../../../_common/form-vue/form.service';
+import { BaseForm, FormOnLoad, FormOnSubmit, FormOnSubmitSuccess } from '../../../../../../_common/form-vue/form.service';
 import { AppImgResponsive } from '../../../../../../_common/img/responsive/responsive';
 import { ModalConfirm } from '../../../../../../_common/modal/confirm/confirm-service';
 
 interface FormModel extends CommunityChannel {
-	_permissions: string;
```
YLivay: Also remove the field on the community channel model.

src/_common/community/channel/channel-permissions.ts
```diff
@@ -0,0 +1,29 @@
+export class ChannelPermissions {
```
hworld: Shouldn't this be called CommunityChannelPermissions?

src/_common/community/channel/channel-permissions.ts
```diff
@@ -0,0 +1,29 @@
...
+	public static readonly ACTION_POSTING = 'posting';
```
hworld: public by default

hworld: I think we should export consts outside of the class, though.
```
export const COMMUNITY_CHANNEL_PERMISSIONS_ACTION_POSTING = 'posting';
```
I know it's long in this case, but it allows components to use the const without having to import the whole model class.

src/_common/community/channel/channel-permissions.ts
```diff
@@ -0,0 +1,29 @@
...
+
+	allPermissions = false;
+	noPermissions = false;
+
+	perms?: { [key: string]: boolean };
```
hworld: I think there's a new-ish Record type in TS: `perms?: Record<string, boolean>` A bit nicer to type.

src/_common/community/channel/channel-permissions.ts
```diff
@@ -0,0 +1,29 @@
...
+			this.perms = input;
+		}
+	}
+
+	public canPerform(action: string) {
+		if (this.noPermissions) {
+			return false;
+		} else if (this.allPermissions) {
+			return true;
+		} else if (this.perms && this.perms[action] === true) {
```
hworld: Prime candidate for safe navigation!
`} else if (this.perms?[action] === true) {`
I think that's the syntax when using [] at least.

src/_common/fireside/post/post-model.ts
```diff
@@ -287,6 +288,21 @@ export class FiresidePost extends Model implements ContentContainerModel, Commen
 		return null;
 	}
 
+	canPublishToCommunities() {
+		if (!this.communities) {
+			return true;
+		}
+		for (const community of this.communities) {
+			if (
+				!community.channel ||
```
hworld: TS safe navigation operator?

src/app/components/forms/community/channel/edit/edit.ts
```diff
@@ -2,39 +2,52 @@ import { Component, Prop } from 'vue-property-decorator';
...
 
 	onLoad(payload: any) {
 		this.maxFilesize = payload.maxFilesize;
 		this.maxWidth = payload.maxWidth;
 		this.maxHeight = payload.maxHeight;
+
+		this.setField('permission_posting', payload.permission_posting ?? 'all');
+	}
+
+	onSubmit() {
```
hworld: Why are we manually doing this work instead of letting the form handle it now?

src/app/components/forms/community/channel/edit/edit.vue
```diff
@@ -47,6 +47,25 @@
...
 
+		<app-form-group name="permission_posting" :label="$gettext('Channel posting permissions')">
+			<div class="help-inline">
+				<span v-translate>
+					Choose who can post to this channel.
+				</span>
+			</div>
+			<div
+				class="radio"
+				v-for="(permissionDisplay, permission) in permissionPostingOptions"
```
hworld: nitpick, but we use `of` instead of `in`

src/app/components/forms/post/post.ts
```diff
@@ -347,6 +348,12 @@ export default class FormPost extends BaseForm<FormPostModel>
 		if (!this.selectedChannel) {
 			this.selectedChannel = this.defaultChannel || null;
 		}
+		if (
+			this.selectedChannel &&
+			!this.selectedChannel.permissions.canPerform(ChannelPermissions.ACTION_POSTING)
```
hworld: safe nav

src/app/components/post/view/view.vue
```diff
@@ -86,6 +86,20 @@
 							/>
 						</div>
 
+						<template v-if="shouldShowCommunityPublishError">
+							<br />
+							<div class="well fill-offset">
+								<app-jolticon icon="notice" notice />
+								<span
+									><translate
+										>This post can currently not be published to the selected community channel
```
hworld: Let's change to:
```
You can't publish this post to the selected community channel because you don't have permissions to post into that specific channel. Please select a different channel.
```

src/app/components/forms/community/channel/edit/edit.ts
```diff
@@ -21,12 +17,10 @@ interface FormModel extends CommunityChannel {
 	},
 })
 export default class FormCommunityChannelEdit extends BaseForm<FormModel>
-	implements FormOnLoad, FormOnSubmitSuccess {
+	implements FormOnLoad, FormOnSubmitSuccess, FormOnSubmit {
 	@Prop(Community)
 	community!: Community;
 
-	modelClass = CommunityChannel;
```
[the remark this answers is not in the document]
nilllzz: Can't use modelClass if the model used for the form doesn't match with the model class. And we use a custom form model.

src/_common/community/channel/channel-permissions.ts
nilllzz: I don't think that's a thing, for direct array access you still need to check if the array is truthy.
`this.perms?[]` is not valid syntax.

src/app/components/forms/community/channel/edit/edit.ts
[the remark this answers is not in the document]
nilllzz: Actually, nevermind. I made the FormModel a class, which can now be used for the modelClass.

PR #440, opened 2020-03-11
Created initialTab prop for comment modals

hapurubokka:
Created initialTab optional prop for comment modals. This will allow us to set the tab sorting to a value other than the default in the comments-store. It is now set to change the sorting to the "You" tab when a comment is created from an event-item.

Also changed how we update the comment-store when deleting your own comment; it should now update the count and totalCount when a comment is either created or deleted, making the counts a little more seamless.

### DIFF
--- a/src/_common/comment/widget/widget.ts
+++ b/src/_common/comment/widget/widget.ts
@@ -109,7 +109,6 @@ export default class AppCommentWidget extends Vue {
 	resourceOwner: User | null = null;
 	perPage = 10;
 	currentPage = 1;
-	tabSet = false;
 
 	collaborators: Collaborator[] = [];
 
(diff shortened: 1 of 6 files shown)


src/_common/comment/widget/widget.ts
```diff
@@ -264,6 +269,14 @@ export default class AppCommentWidget extends Vue {
 			throw new Error(`You need a store set before fetching comments.`);
 		}
 
+		// Check to see if we received 'initialTab' as a prop. If so, we want to setSort()
+		// to filter comments based on that parameter. This allows us to set the comment
+		// sorting to the "You" tab when you leave a comment from an event item.
+		if (this.initialTab && !this.tabSet) {
```
hworld: If you move this into init() then you don't need tabSet, since it'll only trigger on created() or when model changes.

PR #441, opened 2020-03-11
Limiting follow suggestions through user-follow

hapurubokka:
- Implemented suggestion service to prevent spam-requesting a user to follow the user of a post that they liked.
- Added analytics to event-item-controls for when user-follow is shown and when the 'X' is clicked to dismiss it.


src/app/components/event-item/controls/controls.ts
```diff
@@ -114,11 +116,34 @@ export default class AppEventItemControls extends Vue {
...
+			// of time - specified in UserFollowSuggestion.
+			Analytics.trackEvent('user-follow', 'show', 'fireside-post-like-widget');
+			UserFollowSuggestion.doNotSuggest(this.post.user.id);
+		}
+
+		this.shouldShowFollowState = showing;
 
 		// If we're part of the activity feed, synchronize it with that state as well.
 		if (this.item && this.feed) {
+			console.log('item && feed');
```
hworld, line 137: remove

PR #442, opened 2020-03-11
Display follow buttons and user-follow on home feed posts

hapurubokka:
Changed the route info for home feed to display follow buttons and menus for applicable posts.

### DIFF
--- a/src/app/views/home/feed.ts
+++ b/src/app/views/home/feed.ts
@@ -148,17 +148,12 @@ export default class RouteActivityFeed extends BaseRouteComponent {
 			{
 				type: 'EventItem',
 				url: `/web/dash/activity/more/activity`,
+				shouldShowFollow: true,
 				notificationWatermark: feedPayload.unreadWatermark,
 			},
 			feedPayload.items,
 			fromCache
 		);
-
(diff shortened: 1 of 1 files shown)


src/app/views/home/feed.ts
```diff
@@ -154,6 +154,12 @@ export default class RouteActivityFeed extends BaseRouteComponent {
 			fromCache
 		);
 
+		// Stating that we should allow follow buttons
+		// and menus to appear on the home feed.
+		if (this.feed) {
+			this.feed.shouldShowFollow = true;
```
hworld: You can actually put this in the above object, like so:

```
this.feed = ActivityFeedService.routed(
			this.feed,
			{
				type: 'EventItem',
				url: getFetchUrl(this.$route),
				hideCommunity: true,
				hideCommunityChannel: this.channel !== 'featured' && this.channel !== 'all',
				shouldShowFollow: true,
				notificationWatermark: $payload.unreadWatermark,
			},
			$payload.items,
			fromCache
		);
```

That function takes the feed configuration. Also, I don't think you need a comment since the variable name is pretty descriptive.

PR #443, opened 2020-03-12
Responsive-dimensions max-height based on screen height for mobile

hapurubokka:
responsive-dimensions will now check if the screen size is mobile and set the max-height based on the total screen height of the device. This will allow it to ignore resizing images when a keyboard is opened on mobile.

### DIFF
--- a/src/app/components/activity/feed/devlog-post/media/item/item.ts
+++ b/src/app/components/activity/feed/devlog-post/media/item/item.ts
@@ -48,8 +48,12 @@ export default class AppActivityFeedDevlogPostMediaItem extends Vue {
 			return;
 		}
 
+		// If the screen size is considered mobile, we want to treat
+		// the mobile keyboard as if it doesn't exist. Using the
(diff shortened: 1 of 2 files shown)


src/app/components/activity/feed/devlog-post/media/item/item.ts
```diff
@@ -43,6 +43,17 @@ export default class AppActivityFeedDevlogPostMediaItem extends Vue {
 		return this.isActive;
 	}
 
+	get deviceMaxHeight() {
+		if (GJ_IS_SSR) {
+			return;
+		}
+
+		if (Screen.isMobile) {
+			return screen.height * 0.45;
```
hworld: write this as `window.screen`. Since window is the "global" context that any variable attaches to, I try to reference it directly so that it's obvious when we're pulling from window, such as `window.location` and stuff

hworld: Also, add a comment why we need this if!

PR #448, opened 2020-03-19
Media item avg_img_color set as background-color

hapurubokka:
Sets the background-color of a media item to the average color of the item if the item has no transparency.

Also changed some style bindings into computed getters for a cleaner template.


src/_common/media-item/media-item-model.ts
```diff
@@ -42,6 +42,8 @@ export class MediaItem extends Model {
 	mediaserver_url_webm!: string;
 	mediaserver_url_mp4!: string;
 	mediaserver_url!: string;
+	avg_img_color!: string;
```
hworld: I forgot to tell you. This property should be `null | string`. We dont' ahve the data on old media items. You'll have to update the itemStyling() appropriately, to only apply the background color when this property is passed in

PR #458, opened 2020-03-28
Add AppMediaItemBackdrop, display avg_img_color on applicable items

hapurubokka:
Fix issue where `background-color` was attached to `img` tag, making it only appear when the image loads in, effectively doing nothing.

Add `AppMediaItemBackdrop` as a common component, which allows us to wrap an image/fill its container to display the average color of the image while it's loading.

- Required prop
  - `mediaItem!: MediaItem` to get the `avg_img_color` value from it to display behind the image. If there isn't one, set a fallback if there should be one.

- Optional prop
  - `radius?: 'sm' | 'md' | 'lg' | 'full'` will attach a class to the wrapper using the appropriate `rounded-corners()` mixin, or `border-radius: 50%` in the case of 'full'. This component uses `overflow: hidden` by default, so the image itself doesn't need a border radius.


src/_common/game/thumbnail-img/thumbnail-img.vue
```diff
@@ -6,25 +6,27 @@
 		}"
 	>
 		<div class="-inner">
-			<app-jolticon class="-icon" icon="game" />
+			<app-img-backdrop :item="mediaItem" radius="8px">
+				<app-jolticon class="-icon" icon="game" />
```
hworld: Make the icon white now with some sort of opacity so that it blends into the colors of the backdrop.

PR #461, opened 2020-04-01
Stickers improve

nilllzz:


src/_common/sticker/select-modal.ts/select-modal.vue
```diff
@@ -71,6 +80,15 @@
...
 		text-align: center
+		font-weight: bold
+
+	&-rarity
+		&-uncommon
+			color: #1bb804
+		&-rare
+			color: #18a5f2
+		&-epic
+			color: #ffbc56
```
YLivay, line 91: sounds like we'd want to import these variables from some stickers variable file.

PR #464, opened 2020-04-03
Content Mediaex

nilllzz:


src/_common/content/components/media-item/media-item.ts
```diff
@@ -130,6 +132,12 @@ export default class AppContentMediaItem extends Vue {
 		return !this.imageLoaded && !GJ_IS_SSR;
 	}
 
+	get useResponsiveImg() {
```
hworld, line 135: shouldUseReponsiveImg is the convention we use

src/_common/content/components/media-item/media-item.ts
```diff
@@ -130,6 +132,12 @@ export default class AppContentMediaItem extends Vue {
...
+		// For animated media items (gifs):
+		// Always return the direct url because gifs do not get returned by mediaserver.
+		return !!this.mediaItem && !this.mediaItem.is_animated && !!this.mediaItem.mediaserver_url;
```
hworld, line 138: Can't you just do `return !!(this.mediaItem?.is_animated && this.mediaItem?.mediaserver_url)` now?

src/_common/content/components/media-item/media-item.ts
hworld: Also, I actually think this should be changed to `shouldUseMediaserver` since we're using the img-responsive class on both flows. It's more that we want to do the mediaserver flow in certain cases.

PR #465, opened 2020-04-03
Game media mediaserver

nilllzz:


src/_common/content/components/media-item/media-item.ts
```diff
@@ -130,6 +132,12 @@ export default class AppContentMediaItem extends Vue {
 		return !this.imageLoaded && !GJ_IS_SSR;
 	}
 
+	get useResponsiveImg() {
```
hworld, line 135: Looks like we'll have a merge conflict here

src/_common/game/screenshot/screenshot.model.ts
```diff
@@ -11,9 +11,15 @@ export class GameScreenshot extends Model {
...
-	img_thumbnail!: string;
-	img_thumbnail_med!: string;
-	img_thumbnail_large!: string;
+
+	// Needed for lightbox and other components that are used together with GameVideo.
+	get img_thumbnail() {
+		return this.media_item.mediaserver_url;
+	}
+
+	set img_thumbnail(img: string) {
```
hworld, line 20: You don't need the setters for these, right?

nilllzz: We do, through Object.assign in forms we assign this back onto the model.

PR #467, opened 2020-04-06
Get rid of reflect metadata

YLivay:
Very few changes to avoid inferring TS types in some props around the codebase where possible


src/claim/views/key/_bundle/bundle.ts
```diff
@@ -12,13 +13,13 @@ import { Store } from '../../../store/index';
 	},
 })
 export default class AppKeyBundle extends Vue {
-	@Prop()
+	@Prop({ required: true })
```
hworld, line 16: can't this be propRequired(Object)?

src/utils/polyfills.ts
```diff
@@ -1,6 +1,5 @@
-import 'core-js/modules/es7.object.values';
 import 'core-js/modules/es7.object.entries';
-import 'reflect-metadata';
+import 'core-js/modules/es7.object.values';
```
hworld, line 2: why is this line after now?

src/_common/scroll/inview/parent.ts
```diff
@@ -7,7 +7,7 @@ import { ScrollInviewContainer } from './container';
 export class AppScrollInviewParent extends Vue {
 	// If this is a child of AppScrollScroller, we need to get it as a prop so
 	// we can use that scroll element as the root context.
-	@Prop() scroller?: AppScrollScrollerTS;
+	@Prop({ required: false }) scroller?: AppScrollScrollerTS;
```
hworld, line 10: propRequired(Object)

package.json
```diff
@@ -63,7 +63,6 @@
 		"date-fns": "1.30.1",
 		"decompress-zip": "0.3.1",
 		"express": "4.17.1",
-		"extract-loader": "3.0.0",
```
hworld, line 66: Why is this removed in this PR? Just wanted to clean up? Are you 100% sure we don't use it?

src/utils/polyfills.ts
YLivay: automatic import ordering.

package.json
YLivay: yeah, this hasnt been used for ages now.

src/utils/polyfills.ts
hworld: it shouldn't reorder these raw imports....

PR #468, opened 2020-04-06
Remove RxJS

YLivay:
* Implemented our own debouncing
* Made EventBus work as a service and as an instance
* Implemented simple event topic and subscribers on top of new EventBus


src/_common/events/event-bus.service.ts
```diff
@@ -0,0 +1,49 @@
+import Vue from 'vue';
```
hworld, line 1: our folder naming convention is to use singular and have it be the beginning of the class name. So the folder should be "event" not "events"

src/_common/events/event-bus.service.ts
```diff
@@ -0,0 +1,49 @@
...
+	on(event: string, callback: Function): EventBusDeregister {
+		this.e.$on(event, callback);
+
+		return () => {
+			this.e.$off(event, callback);
+		};
+	}
+
+	once(event: string, callback: Function) {
+		this.e.$once(event, callback);
```
hworld, line 27: shouldn't this technically also return a deregister func?

src/_common/events/subscription.ts
```diff
@@ -0,0 +1,26 @@
+export type UnsubscribeFunc = () => void;
+
+export class Subscription {
```
hworld, line 3: Change to EventSubscription

src/_common/events/topic.ts
```diff
@@ -0,0 +1,18 @@
+import { EventBus } from './event-bus.service';
+import { Subscription } from './subscription';
+
+export class Topic<T> {
```
hworld, line 4: EventTopic

src/_common/events/event-bus.service.ts
```diff
@@ -0,0 +1,49 @@
...
+
+	static on(event: string, callback: Function): EventBusDeregister {
+		return this.global.on(event, callback);
+	}
+
+	static once(event: string, callback: Function) {
+		this.global.once(event, callback);
+	}
+
+	static off(event?: string, callback?: Function) {
```
hworld, line 46: let's get rid of the functionality for getting rid of all global events, i think we want this to be as "global" as possible right now. unless we actually do use that in Client maybe?

src/_common/events/subscription.ts
```diff
@@ -0,0 +1,26 @@
...
+	private unsubFunc: UnsubscribeFunc | null;
+
+	constructor(private subscriber: () => UnsubscribeFunc) {
```
hworld, line 6: This unsub logic twisted my head into a pretzel

src/_common/screen/screen-service.ts
```diff
@@ -224,7 +223,8 @@ if (!GJ_IS_SSR) {
 	 * This is used internally to check things every time window resizes.
 	 * We debounce this and afterwards fire the resizeChanges for everyone else.
 	 */
-	fromEvent(window, 'resize')
-		.debounceTime(250)
-		.subscribe(() => Screen._onResize());
+	window.addEventListener(
```
hworld, line 226: We might want to put the debounce logic within Subscription itself so that we can debounce ealier (before it even hits the event bus). The reason is that for stuff like resize, it gets called anytime the window resizes any 1px. We want to make sure it stays super fast if possible.

src/_common/events/event-bus.service.ts
YLivay: event was taken for event item, and this way its similar to the infamous nodejs' events namespace. maybe we need a different place to put these instead of directly under _common?

src/_common/events/event-bus.service.ts
YLivay: I didn't change this bit in event bus. Actually, we don't even use once. I'm going to remove this.

src/_common/events/event-bus.service.ts
YLivay: We don't use it. i'm removing.

src/_common/events/subscription.ts
YLivay: i like pretzels.
anyways the idea is that you can resubscribe to a topic you've previously unsubscribed to. also this way you can't just randomly create a subscription for a topic without going through topic. i hate having many ways to do the same thing.

src/_common/screen/screen-service.ts
YLivay: this is mostly for backwards compat. if we need to react on resize quickly, wouldn't we normally use the resize observer instead?

src/_common/events/event-bus.service.ts
hworld: yeah, I think we should have "system" or something where we put stuff like this

src/_common/screen/screen-service.ts
hworld: I think we need to discuss what I mean with this

src/_common/events/event-bus.service.ts
nilllzz: Why can't we keep it in the 'event-bus' folder?

src/_common/events/event-bus.service.ts
```diff
@@ -0,0 +1,33 @@
+import Vue from 'vue';
+
+export type EventBusDeregister = () => void;
+
+export class EventBus {
+	private e!: Vue;
```
nilllzz, line 6: Please call this something like "element", or even "elem".
With "e" I need to figure out what this even is.

src/_common/events/subscription.ts
nilllzz: I think we need some sort of comment on the EventSubscription or the ctor to explain what's going on here. I can figure it out, but it's kinda convoluted and I'm sure when I actually want to use this, I'll have to go figure it out again.
Some simple explanation of what a subscriber is, what an UnsubscribeFunc is and when/why to use it would be gud.

src/_common/events/subscription.ts
```diff
@@ -0,0 +1,26 @@
...
+	subscribe() {
+		if (!this.unsubFunc) {
+			this.unsubFunc = this.subscriber();
+		}
+
+		return this;
+	}
+
+	unsubscribe() {
+		if (this.unsubFunc) {
```
nilllzz, line 19: Does this mean the unsubFunc is state for whether we are subscribed or not? I think we should have an additional value like isSubscribed or something.

src/_common/events/topic.ts
```diff
@@ -0,0 +1,18 @@
+import { EventBus } from './event-bus.service';
+import { EventSubscription } from './subscription';
+
+export class EventTopic<T> {
+	private e!: EventBus;
```
nilllzz, line 5: Just do `private e = new EventBus();` and remove the constructor function.

nilllzz: Also, rename "e" in the same way as in EventBus.

src/_common/events/event-bus.service.ts
YLivay: Because its no longer just for event bus?..

src/_common/events/subscription.ts
YLivay: I wouldn't say its convoluted, considering thats just the standard naming convention for the components in pubsub systems. I don't mind adding comments tho.

src/_common/events/event-bus.service.ts
YLivay: its self contained within the event bus, i don't see why make it super verbose when its one "goto definition" away.

src/_common/events/subscription.ts
YLivay: what?

src/_common/screen/screen-service.ts
hworld: I was wrong

src/_common/events/subscription.ts
YLivay: removed the resubscribing bit, and actually the entire event subscription. its just a ts type now.

PR #473, opened 2020-04-09
Support deactivating and reusing a custom site domain

YLivay:


src/app/components/sites/manage-page/template.ts
```diff
@@ -28,9 +28,15 @@ export default class AppSitesManagePageTemplate extends Vue {
 	}
 
 	enable() {
-		if (this.site) {
-			return this.site.$activate();
+		if (!this.site) {
+			return;
 		}
+
+		this.site.$activate().catch(e => {
```
nilllzz, line 35: Don't we need to return the promise or resolve it? I think this code currently doesn't do anything.

src/app/components/sites/manage-page/template.ts
```diff
@@ -28,9 +28,15 @@ export default class AppSitesManagePageTemplate extends Vue {
...
+			if (e.errors && e.errors.domain_in_use) {
+				Growls.error('Domain is already in use in another site.');
```
nilllzz, line 37: Forgot translate.

src/app/components/sites/manage-page/static.ts
```diff
@@ -46,9 +46,17 @@ export default class AppSitesManagePageStatic extends Vue {
...
 			const response = await Api.sendRequest(
 				`/web/dash/sites/activate-primary-build/${this.site.id}`,
-				{}
+				{},
+				{ noErrorRedirect: true }
 			);
-			this.site.assign(response.site);
+
+			if (response.errors && response.errors.domain_in_use) {
+				Growls.error(this.$gettext('Domain is already in use in another site.'));
```
nilllzz, line 54: I think we'd need to return after this, otherwise it will still continue.

YLivay: Ah yeah, it used to have some other logic at the end.

src/app/components/sites/manage-page/template.ts
YLivay: Nah, this works as intended. it's not like the html click handler awaits on the promise or anything.

src/app/components/sites/manage-page/template.ts
YLivay: i did

PR #479, opened 2020-04-14
Only display "X NEW ITEMS" in feed when it's in view

hapurubokka:
Added AppScrollInview to the "X NEW ITEMS" button to only display and animate it when you scroll it into view.

### DIFF
--- a/src/app/components/activity/feed/feed.vue
+++ b/src/app/components/activity/feed/feed.vue
@@ -6,18 +6,20 @@
 	-->
 	<div class="activity-feed" :key="feed.id">
 		<template v-if="newCount > 0 || feed.isLoadingNew">
-			<app-expand v-if="!feed.isLoadingNew" when animate-initial>
-				<app-activity-feed-new-button @click="loadNew()">
-					<translate
-						:translate-n="newCount"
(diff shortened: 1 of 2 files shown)


src/app/components/activity/feed/feed.vue
```diff
@@ -6,18 +6,20 @@
...
-						:translate-params="{ count: number(newCount) }"
-						translate-plural="%{count} new items"
-					>
-						1 new item
-					</translate>
-				</app-activity-feed-new-button>
-			</app-expand>
-			<app-loading v-else class="loading-centered" />
+			<app-scroll-inview :margin="`-${Scroll.offsetTop}px`" @inview="onNewButtonInview">
+				<app-expand v-if="!feed.isLoadingNew" :when="isNewButtonInview" animate-initial>
```
hworld, line 10: If you scroll back down, does the space collapse and everything get shifted up?

hapurubokka: Added `isNewButtonInview = false` at the end of the `loadNew()` method, which gets triggered when the button is clicked. This will allow us to reset inview back to its default value, and it should only trigger `inview` as true when the button appears again.

PR #482, opened 2020-04-15
Add text-overflow() to '@username' for forum topic author

hapurubokka:
Added text-overflow() to both parts of the username where we show who made a forum topic.


src/app/components/forum/topic-list/topic-list.styl
```diff
@@ -29,8 +29,10 @@
 			width: 40px
 
 		&-info
-			&-username
+			div
```
hworld, line 32: change this to `> div` otherwise you're targeting every div inside the info, whether it's nested or not

PR #484, opened 2020-04-20
Un-round corners of img in post view on mobile

hapurubokka:
- Removed `AppActivityFeedDevlogPostMediaItem`, added common component `AppMediaItemPost`.
  1. Added optional props: `restrict-device-max-height`, `restrict-should-play-video`, and `inline` to keep functionality of feed view.
- Fixed some bugs in feed view of posts:
  1. Fixed image resizing before `imgloadchange` event triggers from `AppImgResponsive`. This fixes when you would make the browser smaller than the image, and the image would resize before the backdrop.
  2. Fix slider, will now properly resize itself when window size changes. Fixes the image being offset when you resized on any item other than the first.
- Un-round the corners of media items in post view when they span the width of the screen.
- Fix loading background-color corners bleeding from behind image when an image is in an article.


src/app/components/post/view/view.vue
```diff
@@ -19,7 +19,11 @@
 									:max-width="item.width"
 								>
 									<app-event-item-media-tags :gif="item.is_animated" />
-									<app-media-item-backdrop :media-item="item" radius="lg">
+									<app-media-item-backdrop
+										class="-backdrop"
+										:media-item="item"
+										:radius="Screen.isXs ? null : 'lg'"
```
hworld, line 25: We only want the radius removed if it doesn't span the full width.

PR #492, opened 2020-04-27
Update packages, round 1

hapurubokka:
Packages with breaking changes are linked below to their respective version updates.

Changes that needed to be made or items that might affect us are listed below the links.

polished
-------
- [v1.9.2 (current)](https://github.com/styled-components/polished/releases/tag/v1.9.2) -> [v2.0.0](https://github.com/styled-components/polished/releases/tag/v2.0.0) -> [v3.0.0](https://github.com/styled-components/polished/releases/tag/v3.0.0)
- `mix` now requires a ratio to work. We already use ratios when we use mix, so this shouldn't be an issue.

date-fns
--------
- [1.30.1 (current)]() -> [2.0.0 (lots of changes)](https://github.com/date-fns/date-fns/releases/tag/v2.0.0)
- [Too many breaking changes to list, but I looked over them all and tested everything I could.](https://github.com/date-fns/date-fns/blob/master/CHANGELOG.md#200---2019-08-20)


src/_common/filters/date.ts
```diff
@@ -10,20 +10,21 @@ import * as format from 'date-fns/format';
...
 };
 
 export function date(now: Date | string | number, formatTo?: string) {
 	if (formatTo && (ShortcutFormats as any)[formatTo]) {
 		formatTo = (ShortcutFormats as any)[formatTo];
 	}
 
-	return format(now, formatTo);
+	// review, format doesn't accept string anymore.
+	return format(now as number, formatTo as string);
```
hworld, line 29: Check if we use now as a string ever intstead. Modify all cases that are passing in a string.
For the formatTo either set a default value in the function argument, or check if we have any cases where we call this without formatTo. If not, we can just make it required.

hapurubokka: I set `ShortcutFormats.medium` as the default for `formatTo`, just in case anyone forgets to set a format in the future.

src/_common/filters/date.ts
```diff
@@ -20,11 +20,11 @@ const ShortcutFormats = {
 	shortTime: 'h:mm a',
 };
 
-export function date(now: Date | string | number, formatTo?: string) {
+export function date(now: Date | number, formatTo: string = ShortcutFormats.medium) {
```
hworld, line 23: No need to specify string for formatTo, it'll infer it from the type of "medium"

PR #493, opened 2020-04-28
Content media item filesize

nilllzz:


src/_common/content/components/media-upload/media-upload.ts
```diff
@@ -111,6 +106,43 @@ export default class AppContentMediaUpload extends Vue {
 		);
 		if ($payload.success && $payload.mediaItems && $payload.mediaItems.length === 1) {
 			return new MediaItem($payload.mediaItems[0]);
+		} else if ($payload.success === false && $payload.errors.file) {
```
YLivay, line 109: `!$payload.success`, I'm not sure if we consistently return false or maybe just not return success at all in some cases.

src/_common/content/components/media-upload/media-upload.ts
```diff
@@ -111,6 +106,43 @@ export default class AppContentMediaUpload extends Vue {
...
+				}
+			);
+
+			const maxWidth = sizePayload.maxWidth;
+			const maxHeight = sizePayload.maxWidth;
+
+			Growls.error({
+				title: this.$gettext('Oh no!'),
+				message: this.$gettextInterpolate(
+					"It looks like your image's filesize or dimensions are too large. The maximum allowed size is %{ width }x%{ height } pixels.",
```
YLivay, line 127: lets also output the max filesize.

nilllzz: Yeah. I'll add the parameter, but @hworld has to figure out the message.

hworld: It looks like your image's filesize or dimensions are too large. Its filesize must be less than %{filesize} and its dimensions less than NxN.

PR #495, opened 2020-04-30
Better reporting

nilllzz:


src/_common/form-vue/form.service.ts
```diff
@@ -91,6 +91,15 @@ export class BaseForm<T> extends Vue {
...
 	get valid() {
+		return this._isValid();
+	}
+
+	/**
+	 * This exists so we have a backing field in case we want to override valid()
+	 * and want to extend its function body.
+	 * Since you cannot invoke super getters directly, call this function instead of `super.valid`.
+	 */
+	protected _isValid() {
```
YLivay, line 102: im on the fence with this but @hworld will probably be more opinionated.

I think we should just mutate customErrors if we have any additional logic that needs to be handled by the form.

src/_common/report/form/form.ts
```diff
@@ -1,22 +1,309 @@
...
+		}
+
+		// Check that if we have contexts to choose from, at least one of them is selected.
+		if (!this.formModel.reason) {
+			return true;
+		}
+		const reason = this.reasons.find(i => i.radioValue === this.formModel.reason);
+		if (!reason) {
+			return true;
+		}
```
YLivay, line 43: we should probably be throwing here, yeah? theres no such thing as a non existing reason object.

src/_common/report/form/form.ts
```diff
@@ -1,22 +1,309 @@
...
+
+				// For a devlog post of a game that is maturity restricted, we don't want to show the "explicit" report option.
+				// Those devlog posts can be explicit, and we don't want to encourage false reports.
+				if (
+					this.resource instanceof FiresidePost &&
+					this.resource.game instanceof Game &&
+					this.resource.game.tigrs_age === 3
+				) {
+					arrayRemove(reasons, i => i.radioValue === 'explicit');
+				}
```
YLivay, line 137: you're assuming game posts only show on the game page.
this might be changing very soon as you know (wink wink).
consider reverting this.

nilllzz: I'd argue if we allow showing the post somewhere else, we can add that check here.
Only show this option, if the post is solely visible on the game page.

PR #499, opened 2020-05-04
Remove simplebar, use css for scrollbars instead

hapurubokka:
We're able to style scrollbars using `scrollbar-width` and `scrollbar-height` on Firefox, and using webkit pseudo-selectors for other browsers.

- [Firefox documentation](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Scrollbars)
- [Chrome, Edge, Safari, etc. documentation](https://developer.mozilla.org/en-US/docs/Web/CSS/::-webkit-scrollbar)

This PR removes everything related to simplebar and instead uses the props through AppScrollScroller to modifying the styling of the native scrollbar.

AppScrollScroller will now color the scrollbar anywhere it's used, but optional props are:
1. `thin?: boolean` will change the width of the scrollbar
2. `hideScrollbar?: boolean` will change the display of the scrollbar to hidden.
3. `horizontal?: boolean` will specify that the content is intended to only overflow horizontally.


src/_common/comment/widget/comment/comment.vue
```diff
@@ -23,7 +23,7 @@
 		</template>
 
 		<template #meta>
-			<app-popper v-if="user">
+			<app-popper v-if="user" popover-class="list-group-dark">
```
hworld, line 26: Can we instead just use the correct "fill-X" class? We have the list-group-dark in the popover slot below, so I think we just need this to change the BG to the correct color. I'm guessing list-group-dark is the same as fill-dark, but I don't know.

hworld: And then be sure to update all the places doing the same thing.

src/_common/media-bar/media-bar.ts
```diff
@@ -19,7 +19,7 @@ import AppMediaBarLightbox from './lightbox/lightbox.vue';
 })
 export default class AppMediaBar extends Vue {
 	@Prop(Array) mediaItems!: any[];
-	@Prop(Boolean) noOverlayScrollbars?: boolean;
+	@Prop(Boolean) thinScrollbar?: boolean;
```
hworld, line 22: I think you can get rid of this prop and just always style it as thin. I think we only had this for game jolt sites, which should now work fine with our normal scrollbar styling. We will just have to remember to check a game jolt site when we push.

src/_common/scroll/scroller/scroller.ts
```diff
@@ -10,65 +9,24 @@ import { AppScrollInviewParent } from '../inview/parent';
 	},
 })
 export default class AppScrollScroller extends Vue {
-	@Prop(Boolean)
-	overlay?: boolean;
+	@Prop(propOptional(Boolean))
```
hworld, line 12: Change these to: `@Prop(propOptional(Boolean, false)) thin: boolean` This way we dont' ahve undefined.

src/_common/scroll/scroller/scroller.ts
```diff
@@ -10,65 +9,24 @@ import { AppScrollInviewParent } from '../inview/parent';
...
-		return this.overlay && !GJ_IS_SSR;
-	}
-
-	/**
-	 * The actual Element that scrolls for this Scroller.
-	 */
-	get scrollElement() {
-		return this._scrollElement;
-	}
+	_scrollElement!: HTMLElement;
```
hworld, line 22: You can actually get rid of this too, just keep the isMounted. Then in the template you can just pass in $el to the child prop. We just need the ismount to make sure we render it when we're mounted.

src/_common/scroll/scroller/scroller.vue
```diff
@@ -2,61 +2,64 @@
...
-					</app-scroll-inview-parent>
-				</div>
-			</div>
-		</template>
-		<template v-else>
-			<app-scroll-inview-parent v-if="isMounted" :scroller="this">
-				<slot />
-			</app-scroll-inview-parent>
-		</template>
+		<app-scroll-inview-parent v-if="isMounted" :scroller="_scrollElement">
```
hworld, line 10: Basically this can become
```
<app-scroll-inview-parent v-if="isMounted" :scroller="$el">
```

src/_common/scroll/scroller/scroller.vue
```diff
@@ -2,61 +2,64 @@
...
+				border-radius: $-thumb-radius
+
+	&.-hide-scrollbar
+		scrollbar-width: none
+
+		&::-webkit-scrollbar
+			display: none
+
+	// Making sure that we don't have transparency issues with the track in modals
+	&.modal
```
hworld, line 58: We need to improve the thumb color in light mode.

PR #502, opened 2020-05-07
Enlarge media

nilllzz:
Provides ways to show larger versions of images in fireside posts.


src/_common/media-bar/lightbox/item/item.vue
```diff
@@ -18,13 +18,22 @@
...
 							:alt="item.caption"
 							:style="{
 								width: 'inherit',
 								height: 'inherit',
 							}"
 						/>
+						<app-video
+							v-else
+							class="-video"
+							:poster="item.getMediaItem().mediaserver_url"
```
hworld, line 32: Put these templates expressions that call functions into vue getters.

PR #510, opened 2020-05-22
Add tooltip touch support through 'touchable' modifier

hapurubokka:
Add a `touchable` modifier to the AppTooltip directive to allow tooltips to display on touch events. Then we can allow tooltips on mobile for our `?` icons.

### DIFF
--- a/src/_common/popper/popper.ts
+++ b/src/_common/popper/popper.ts
@@ -183,6 +183,13 @@ export default class AppPopper extends Vue {
 		Popper.registerPopper(this.$router, this);
 	}
 
+	beforeDestroy() {
+		// We have to remove open poppers if their reference gets destroyed.
+		if (this.isVisible) {
+			document.body.removeChild(this.$refs.popper);
+		}
+	}
+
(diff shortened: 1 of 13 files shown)


src/_common/popper/popper.ts
```diff
@@ -300,6 +298,23 @@ export default class AppPopper extends Vue {
 		document.addEventListener('click', this.onClickAway, true);
 	}
 
+	private destroyPopper() {
+		this.clearHideTimeout();
+
+		// Clean up any remaining popper elements and instances
+		if (document.body.contains(this.$refs.popper)) {
```
hworld, line 305: You might be able to just do this.$refs.popper.remove() so that we don't ahve to check the body.contains. I'm just worried that the contains call could be slow: https://developer.mozilla.org/en-US/docs/Web/API/ChildNode/remove Thoughts?

hapurubokka: That sounds cleaner to me. Changed to check if `this.$refs.popper` exists, and if so then call `remove()`.

PR #519, opened 2020-05-30
Fix user header menu wrapping and breaking content

hapurubokka:
Change from using `col-lg-X` styling to using flex so we can shrink the controls if needed.

This will fix the content breaking on profiles that wrap the nav when there is too much information.
To do this it means that the control button(s), like 'Follow' or 'Edit', must shrink to make space if needed until they reach a breakpoint that puts them on a different line (< 1200px).

1200px wide with high counts for the nav buttons
image

### DIFF
--- a/src/app/components/page-header/page-header.styl
+++ b/src/app/components/page-header/page-header.styl
@@ -3,6 +3,24 @@
 
 $-cover-buttons-height = $button-md-line-height + 20px
 
+.-index
+	.-controls, .-nav
+		padding: 0 20px
+
+	@media $media-xs
+		.-controls, .-nav
+			padding: 0 15px
+
+	@media $media-lg-up
+		display: flex
+		flex-direction: row-reverse
+		justify-content: flex-end
+
+		.-controls
+			max-width: 33.333333%
+			flex: auto
+			margin-left: auto
+
 .page-header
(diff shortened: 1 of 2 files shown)


src/app/components/page-header/page-header.styl
```diff
@@ -3,6 +3,24 @@
 
 $-cover-buttons-height = $button-md-line-height + 20px
 
+.-index
+	.-controls, .-nav
+		padding: 0 20px
```
hworld, line 8: This needs to use $grid-gutter-width

src/app/components/page-header/page-header.styl
```diff
@@ -3,6 +3,24 @@
...
+
+	@media $media-xs
+		.-controls, .-nav
+			padding: 0 15px
```
hworld, line 12: $grid-gutter-width-xs

src/app/components/page-header/page-header.styl
hapurubokka: changed to `padding: 0 ($grid-gutter-width / 2)`

src/app/components/page-header/page-header.styl
hapurubokka: changed to `padding: 0 ($grid-gutter-width-xs / 2)`

PR #523, opened 2020-06-04
YouTube: No channel error message

nilllzz:
Handle new "no-channel" error message from backend (and update formatting).


nilllzz:
> @nilllzz pls check your max width setting for prettier. it should be 100 as specified in the project's .prettierrc.
> 
> how are we getting these diffs every time

Checked it, it's set to 100. That is why this file was formatted the way it is. Should I not commit the reformat and instead just the change?

PR #530, opened 2020-06-07
Less posts per page

nilllzz:
We will start to return only 10 posts per feed page instead of 15.
Since we still return 15 notifications per page, we need to be able to control how many items per page we expect based on the feed.

We also increase the auto scroll limit from 3 to 5, for a total post auto load of 50 instead of 45.


nilllzz:
Superseded by https://github.com/gamejolt/gamejolt/pull/661

PR #531, opened 2020-06-08
Super naive reconnection for chat

YLivay:
This is a temporary stopgap.

### DIFF
--- a/src/app/components/chat/client.ts
+++ b/src/app/components/chat/client.ts
@@ -155,9 +155,8 @@ function reconnect(chat: ChatClient) {
 }
 
 async function connect(chat: ChatClient) {
-	const chatId = ChatClient.nextId;
+	const chatId = ChatClient.nextId++;
 	chat.id = chatId;
-	ChatClient.nextId = ChatClient.nextId++ % 100;
 
 	const frontend = await getCookie('frontend');
 	const user = store.state.app.user;
--- a/src/app/components/chat/client.ts
+++ b/src/app/components/chat/client.ts
@@ -28,7 +28,13 @@ export interface ChatNewMessageEvent {
  * @param requestGetter Function that generates a promise that represents the
  * request
  */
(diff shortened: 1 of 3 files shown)


hworld:
@ttbowen and @nilllzz can you review this? It's important that we get reconnects fixed up.

src/app/components/chat/client.ts
```diff
@@ -157,18 +182,29 @@ function reset(chat: ChatClient) {
 }
 
 async function connect(chat: ChatClient) {
+	const chatId = ChatClient.nextId;
+	chat.id = chatId;
+	ChatClient.nextId = (ChatClient.nextId + 1) % 100;
```
nilllzz, line 187: May I ask why?

src/app/components/chat/client.ts
```diff
@@ -193,7 +229,21 @@ async function connect(chat: ChatClient) {
...
+		console.warn(e);
+		chat.reconnect();
+	});
+
+	chat.socket.onClose(() => {
+		console.warn('[Chat] Socket closed unexpectedly');
+		chat.reconnect();
+	});
+
+	chat.socket.sendHeartbeat();
```
nilllzz, line 243: the @types seem to not have this function defined, but phoenix's js also doesn't have this marked as private, so I guess this is valid?

nilllzz: Think we just need a quick comment since this is marked as an error in TS

src/app/components/chat/client.ts
YLivay: tracking the chat id as it is reinstantiated.
if it changes, thats how i super hackily abort the rest of the connection flow and the pollRequests in the now outdated instance.

src/app/components/chat/client.ts
YLivay: oh i don't think i should've commented the sendHeartbeat, i was debugging with it, is all.

PR #534, opened 2020-06-09
Only show post fullscreen button on larger items

hapurubokka:
other fixes:
- lightbox items will no longer vertically shift if adjacent content is smaller than 100% of the containers content area height
- moved `dragging` class assignment - dragging lightbox items will now properly set `cursor: grabbing`

### DIFF
--- a/src/_common/media-item/post/post.ts
+++ b/src/_common/media-item/post/post.ts
@@ -52,8 +52,8 @@ export default class AppMediaItemPost extends Vue {
 	get shouldShowFullscreenOption() {
 		return (
 			this.restrictDeviceMaxHeight &&
-			this.mediaItem.height >= 150 &&
-			this.mediaItem.width >= 150
+			this.mediaItem.height >= 100 &&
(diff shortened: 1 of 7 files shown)


src/_common/media-item/post/post.ts
```diff
@@ -49,6 +49,10 @@ export default class AppMediaItemPost extends Vue {
 	@Emit('fullscreen')
 	emitFullscreen(_mediaItem: MediaItem) {}
 
+	get shouldShowFullscreen() {
```
hworld, line 52: We should almost always show fullscreen option, I think. We should just not show it if the media item height or width is less than 100px or something.

hworld: Also, maybe call it "shouldShowFullscreenOption"?

hapurubokka: Changed it to check if the mediaItem height and width is >= 150, and the device height is set to be restricted (currently only used in feed view).

PR #536, opened 2020-06-11
Fetch and pass chat token instead of frontend

ttbowen:
Now using the `chatex/token` to generate a token. This token gets passed to the socket connection instead of the frontend cookie.

### DIFF
--- a/src/_common/environment/environment.service.ts
+++ b/src/_common/environment/environment.service.ts
@@ -41,6 +41,7 @@ export class Environment {
 	static gameserverApiHost = 'https://gamejolt.net';
 	static activityStreamHost = 'https://activity.gamejolt.com';
 	static chatHost = 'https://chatex.gamejolt.com/chatex/host';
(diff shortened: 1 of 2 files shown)


src/_common/environment/environment.service.ts
```diff
@@ -41,6 +41,7 @@ export class Environment {
 	static gameserverApiHost = 'https://gamejolt.net';
 	static activityStreamHost = 'https://activity.gamejolt.com';
 	static chatHost = 'https://chatex.gamejolt.com/chatex/host';
+	static chatToken = 'https://chatex.gamejolt.com/chatex/token';
```
hworld, line 44: Can you instead refactor the code to have one single entry for chat? Just call it chat and assign it "https://chatex.gamejolt.com/chatex" and then in the code that uses it add the /host when needed and /token when needed. It's supposed to be a path prefix, not the full routes.

PR #541, opened 2020-06-16
Toggle post author controls instead of disabling

hapurubokka:
Also sets "Post to profile" to true on default

### DIFF
--- a/src/app/components/forms/post/post.ts
+++ b/src/app/components/forms/post/post.ts
@@ -354,15 +354,15 @@ export default class FormPost extends BaseForm<FormPostModel>
 	}
 
 	@Watch('formModel.post_to_user_profile')
-	onPostToUserProfileToggle(event: boolean) {
-		if (event) {
+	onPostToUserProfileToggle() {
+		if (!!this.formModel.post_to_user_profile) {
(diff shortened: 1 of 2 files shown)


src/app/components/forms/post/post.ts
```diff
@@ -353,6 +353,20 @@ export default class FormPost extends BaseForm<FormPostModel>
 		return this.defaultCommunity;
 	}
 
+	@Watch('formModel.post_to_user_profile')
+	onPostToUserProfileToggle(event: boolean) {
```
hworld, line 357: no need to get the value here, just use this.formModel.post_to_user_profile directly in the function body. It'll be up to date and this way the typing is synced properly.

src/app/components/forms/post/post.ts
```diff
@@ -353,6 +353,20 @@ export default class FormPost extends BaseForm<FormPostModel>
...
+		if (event) {
+			this.setField('as_game_owner', false);
+		}
+	}
+
+	@Watch('formModel.as_game_owner')
+	onPostAsGameOwnerToggle(event: boolean) {
```
hworld, line 364: same with this one

PR #542, opened 2020-06-16
Add multiple-sticker redemption

hapurubokka:


src/_common/modal/modal.ts
```diff
@@ -40,6 +40,8 @@ export default class AppModal extends Vue {
 		scroller: AppScrollScrollerTS;
 	};
 
+	@Emit('click-away') emitClickAway() {}
```
nilllzz, line 43: I think we want to instead call this event "close" and emit it whenever the modal closes:
 - esc button pressed
 - backdrop clicked

src/_common/sticker/collect/collect.ts
```diff
@@ -1,15 +1,23 @@
...
 import { Growls } from '../../growls/growls.service';
+import { Ruler } from '../../ruler/ruler-service';
 import { AppTooltip } from '../../tooltip/tooltip-directive';
 import AppStickerCard from '../card/card.vue';
 import AppStickerCardHidden from '../card/hidden/hidden.vue';
 import { Sticker } from '../sticker.model';
 
+// Sync these up with '../card/variables.styl' - needed to calculate
+// how many cards to show per row when redeeming multiple stickers.
+const CardWidth = 150;
```
nilllzz, line 18: Make consts UPPER_CASE, so `CARD_WIDTH` for example.

src/_common/sticker/collect/collect.vue
```diff
@@ -3,38 +3,70 @@
...
+						</app-sticker-card>
+					</div>
 				</div>
+
+				<div
+					v-if="limitedStickerDisplay.length < purchasedStickers.length"
+					class="-revealed-controls-load-more page-cut"
+				>
+					<app-button trans @click="shownRows++">
+						<translate>Load More</translate>
```
nilllzz, line 46: I think this should say "Show More". I also think it'd be best to just show all when you click this button once, instead of steadily increasing row count.

nilllzz: In that case, we could also display "Show All".

src/_common/modal/modal.ts
hapurubokka: We were already emitting under those circumstances, so I'll change the naming to reflect that.

src/_common/sticker/collect/collect.vue
hapurubokka: I changed it to 'Show All" as requested, but I still have it showing one row every purchase - so it will need to be clicked every time if you want to see them all.

src/_common/modal/modal.ts
```diff
@@ -109,7 +109,7 @@ export default class AppModal extends Vue {
 	}
 
 	dismiss() {
-		this.emitClickAway();
+		this.emitClose();
```
nilllzz, line 112: Also emit on esc key and you are gucci.

hapurubokka: I believe it currently is, but only when `esc` is intended to close the modal.
```
dismissEsc() {
	if (this.modal.noEscClose) {
		return;
	}
	this.dismiss();
}
```
Currently `dismissRouteChange()`, `dismissEsc()`, and `dismissBackdrop()` all close the modal by triggering the `dismiss()` function.

nilllzz: Ah right, I thought it called resolve.

PR #543, opened 2020-06-17
Fix lightbox swipes breaking with iframes

hapurubokka:
This PR allows lightbox dragging to be triggered on all directions, with no threshold so it adds the `dragging` class immediately.

This sets `pointer-events: none` on iframes in lightbox when the `dragging` class exists, fixing the `panend` event not triggering because the iframe can't bubble events back to our root document.

### DIFF
--- a/src/_common/lightbox/lightbox.styl
+++ b/src/_common/lightbox/lightbox.styl
@@ -9,8 +9,9 @@ $-controls-padding = 10px
 	.dragging &
 		cursor: grabbing
 
-		// iframe elements don't allow events to bubble back to the root document, so we need to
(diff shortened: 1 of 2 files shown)


src/_common/lightbox/lightbox.styl
```diff
@@ -6,9 +6,12 @@ $-button-size = 90px
 $-controls-padding = 10px
 
 .media-bar-lightbox
-	&.dragging
+	.dragging &
 		cursor: grabbing
 
+		>>> iframe
```
hworld, line 12: add a comment for why it's needed

hapurubokka: added comment

PR #550, opened 2020-06-24
Fix stickers overlapping fireside-post controls

hapurubokka:


nilllzz:
Latest commit looks good as well, but I found another one:
image
The "Fullscreen" overlay button should also have a high enough z-index to overlay the stickers below it.

hapurubokka:
> Latest commit looks good as well, but I found another one:
> image
> The "Fullscreen" overlay button should also have a high enough z-index to overlay the stickers below it.

I think the only way to fix this is to move those toolbar controls out of the slider element, since the slider element is creating a new stacking context through the `transform` property. This will change the placement of the toolbar element in a few different cases though, like on multi-item posts and items that have their height capped.

It would be nice to put these controls above the stickers, but I think that fixing this z-indexing isn't worth the drawbacks.
image

PR #556, opened 2020-06-29
Make it easier to tell when a webkit scrollbar is hovered

hapurubokka:
Make it easier to see that you're hovering webkit scrollbars.


src/_common/theme/theme.ts
```diff
@@ -106,6 +106,11 @@ export class AppTheme extends Vue {
 				${genDarkVar('link', theme.darkHighlight_)}
 				${copyDarkVar('link-hover', 'white')}
 			}
+
+			.scroll-scroller {
```
hworld, line 110: I think instead of making this a global thing of "themes" we should do it just within the scroller component. We shouldn't be setting scroller styling in the theme component. Copy the theme stuff into the scroller component instead.

src/_common/theme/theme.model.ts
```diff
@@ -302,6 +306,14 @@ export class Theme extends Model {
 		return this.tintColor(GraySubtle, 0.04);
 	}
 
+	get graySubtleHover_() {
```
hworld, line 309: Just do these calculations in the scroller component.

src/_common/theme/theme.model.ts
```diff
@@ -20,6 +20,10 @@ const GrayDarker = '#1f1f1f';
 const GrayDark = '#292929';
 const Gray = '#363636';
 const GraySubtle = '#444444';
+// The value that FireFox scrollbars change to on hover
+const GraySubtleHover = '#565656';
```
hworld, line 24: put these in the scroller component

PR #564, opened 2020-07-02
Modify styling for chat messages

hapurubokka:
- remove rounded corners for messages, bleed to the left edge of the chat window
- highlight whole message and show time when hovering any part of a message
- remove transitions for time and background-color showing


nilllzz:
The new styling for the message hover looks great, but on mobile the avatar alignment is a bit odd:
image
Seems like it's missing some padding on the left, and has too much padding on the right. Would be neat to have it be centered between the left border and the message/username.

PR #572, opened 2020-07-07
Content Editor: Exit inline code

nilllzz:
Adds a command that allows to exit an inline code text node at the end of a block.
This command gets activated when pressing `Right Arrow` or `Space` at the end of an inline code text node which itself is placed at the end of a block node.
It only exits on space when the last character of the code text node is already a space (meaning you have to press space twice to exit when no space character is present.)


src/_common/content/content-editor/plugins/commands/exit-inline-code-command.ts
```diff
@@ -0,0 +1,59 @@
...
+			if (!text || !text.endsWith(' ')) {
+				return false;
+			}
+		}
+
+		// Insert a space at the end of the current block.
+		const tr = state.tr;
+		tr.insertText(' ', state.selection.from);
+		tr.removeMark(state.selection.from, state.selection.from + 1, schema.marks.code);
+		dispatch(tr);
```
YLivay, line 54: If you exit the code by double pressing space, i wonder if we should actually remove the last space in the inline code block.

hapurubokka: I think it would be good to remove that trailing space in the code-text. I feel like that would be the expected behavior for escaping with a double-space, otherwise you'll have to go back into the code-text to remove the space manually.

PR #574, opened 2020-07-09
Comment permalink same page

nilllzz:
Adds a watcher to the comment modal that allows views to watch for route changes that add/change the comment permalink hash.
This watcher is then used in game/user/post views to watch for changes.

### DIFF
--- a/src/_common/comment/thread/modal.service.ts
+++ b/src/_common/comment/thread/modal.service.ts
@@ -47,4 +47,13 @@ export class CommentThreadModal {
 
 		CommentThreadModal.show({ commentId: id, model, displayMode });
 	}
+
+	static watchForPermalink(router: VueRouter, model: Model, displayMode: DisplayMode) {
+		const checkPath = router.currentRoute.path;
+		return router.afterEach((to, _from) => {
+			if (checkPath === to.path && !!to.hash) {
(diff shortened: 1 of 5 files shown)


src/_common/comment/thread/modal.service.ts
```diff
@@ -47,4 +47,13 @@ export class CommentThreadModal {
 
 		CommentThreadModal.show({ commentId: id, model, displayMode });
 	}
+
+	static watchForPermalink(router: VueRouter, model: Model, displayMode: DisplayMode) {
+		const checkPath = router.currentRoute.path;
+		return router.afterEach((to, _from) => {
```
hworld, line 53: This never gets unbound, meaning your event will stick around forever.

src/app/views/discover/games/view/devlog/view/view.ts
```diff
@@ -69,6 +69,8 @@ export default class RouteDiscoverGamesViewDevlogView extends BaseRouteComponent
 
 	post: FiresidePost | null = null;
 
+	permalinkWatcher?: Function;
```
YLivay, line 72: I think we should rename these to make it clearer this IS the deregister. the current name is misleading.

PR #575, opened 2020-07-09
Paginate friend requests

nilllzz:


src/app/components/shell/friend-request-popover/friend-request-popover.ts
```diff
@@ -46,27 +51,68 @@ export default class AppShellFriendRequestPopover extends Vue {
...
+			});
+			this.requestCount = payload.requestCount;
+			this.pendingCount = payload.pendingCount;
+			this.perPage = payload.perPage;
+			this.isBootstrapped = true;
+
+			this.setFriendRequestCount(this.requestCount);
+		}
+
+		this.loadTab();
```
YLivay, line 86: Not awaiting on this means the popover is considered loaded before the first page actually loads. Is this intentional?

nilllzz: No, good catch!

PR #584, opened 2020-07-18
Small no comments/shouts illustration.

hworld:
Also removed the bio on profile pages if there is none, to take up less space.


hworld:
@hapurubokka The reason it's not done is because I don't want to show it if the comments are loading in. We will have to work with the comments store to get that info.

PR #585, opened 2020-07-20
Fix [message] button on profile page for mobile

ttbowen:


ttbowen:
> Nils caught that this is probably a bug, actually. We don't want to load the last active chatroom when you open the friends list on mobile. Will this end up doing that on mobile?

Yes it will.

ttbowen:
> The current changes are still broken. If you're on mobile and you join the chatroom of User_A, change your chatroom to User_B through the friends list, then click on the [MESSAGE] button on User_A's profile, it causes you to join the chatroom of User_B.

You're right. Nice catch!

hapurubokka:
> > The current changes are still broken. If you're on mobile and you join the chatroom of User_A, change your chatroom to User_B through the friends list, then click on the [MESSAGE] button on User_A's profile, it causes you to join the chatroom of User_B.
> 
> You're right I was using the wrong room id. Nice catch!

That seemed to fix the issue for the Xs breakpoint, but caused a new issue with the other breakpoints. If you click [MESSAGE] for User_A, it'll open their chat, but if you then change rooms to User_B and close out of the menu, click [MESSAGE] on User_A again, it'll open the chat for User_A briefly before opening the chat for User_B.

PR #586, opened 2020-07-20
Fix chat notifications not triggering if you have a room open but the window is unfocused.

ttbowen:


src/app/components/chat/notification-growl/notification-growl.service.ts
```diff
@@ -1,13 +1,12 @@
...
-import { ChatClient, enterChatRoom, isInChatRoom } from '../client';
+import { ChatClient, enterChatRoom } from '../client';
 import { ChatMessage } from '../message';
 import AppChatNotificationGrowl from './notification-growl.vue';
 
 export class ChatNotificationGrowl {
 	static async show(chat: ChatClient, message: ChatMessage) {
-		// Skip if already in the room.
-		if (isInChatRoom(chat, message.room_id)) {
+		if (chat.isFocused) {
```
hworld, line 9: I think we need to check their chat room status as well.

PR #588, opened 2020-07-20
Change the styling of chat window controls

hapurubokka:
# xs breakpoint
image
image
# sm breakpoint
image


src/_common/form-vue/control/content/content.ts
```diff
@@ -35,6 +35,8 @@ export default class AppFormControlContent extends BaseFormControlTS {
 
 	@Prop(Object) tempResourceContextData?: Object;
 
+	@Prop(propOptional(Boolean, false)) sleek!: boolean;
```
hworld, line 38: maybe compact would be better

hapurubokka: changed the prop and class

PR #589, opened 2020-07-20
Site guidelines

hworld:


src/_common/comment/add/add.vue
```diff
@@ -25,24 +20,35 @@
 			<app-form-control-errors label="comment" />
 		</app-form-group>
 
-		<div class="text-right">
+		<p v-if="shouldShowGuidelines" class="-guidelines">
+			Remember to be respectful and follow our
```
hworld, line 24: Couldn't translate because it contains a vue component.

nilllzz: Bah and I butchered my community activity. Why can't I just do this there D:

PR #591, opened 2020-07-21
Integrate eslint with TS and vue file support.

hworld:


hworld:
Whatever, I approve of it.

PR #594, opened 2020-07-26
User timeouts

nilllzz:


src/_common/store/app-store.ts
```diff
@@ -34,6 +35,11 @@ export class AppStore extends VuexStore<AppStore, Actions, Mutations> {
 	consents: UserConsents = {};
 	error: number | string | null = null;
 	hasNewStickers = false;
+	timeout: UserTimeout | null = null;
+
+	get isUserTimedOut() {
+		return this.userBootstrapped && !!this.user && !!this.timeout && !this.timeout.isExpired;
```
YLivay, line 41: Just making sure, the timeout property is cleared in backend after the timeout is expired rather than returning it with it's past expiration date?

This should be used purely for handling timeouts expiring without a refresh, ye?

src/_common/store/app-store.ts
```diff
@@ -42,6 +48,19 @@ export class AppStore extends VuexStore<AppStore, Actions, Mutations> {
 		} else {
 			this.user = user;
 		}
+
+		if (user.timeout) {
+			const timeout = new UserTimeout(user.timeout);
+			if (timeout.isExpired) {
+				this.timeout = null;
```
YLivay, line 55: Would this ever happen? imo the backend should not return an expired timeout and we shouldn't care about time drifting.

src/_common/user/timeout/timeout.model.ts
```diff
@@ -0,0 +1,12 @@
+import { Model } from '../../model/model.service';
+
+export class UserTimeout extends Model {
+	expires_on!: number;
+	reason!: string;
+
+	get isExpired() {
+		return this.expires_on <= Date.now();
+	}
```
YLivay, line 9: Is this actually reactive? I think the way to do this is to have an actual Node timer update a property and have the getter use that.

YLivay (changes requested):
Surface level review, overall looks gud, i'll check out the branch after backend.

PR #596, opened 2020-07-27
Track homepage view as '/discover' when no user

hapurubokka:


src/_common/analytics/analytics.service.ts
```diff
@@ -52,6 +53,13 @@ export class Analytics {
 		});
 
 		EventBus.on('routeChangeAfter', () => {
+			// If the home route is being visited and the user is not signed in,
+			// we want to track the event as '/discover' instead of the user feed route.
+			if (router.currentRoute.fullPath === '/' && !this.appUser) {
```
YLivay, line 58: any specific reason to use fullPath and not just path?
fullPath seems to include the query parameters and hash.

hapurubokka: `path` does seem more correct here then - changed

hworld (approved):

PR #603, opened 2020-08-04
Fix post author controls overwriting draft settings

hapurubokka:
Fixes https://github.com/gamejolt/issue-tracker/issues/1402

### DIFF
--- a/src/app/components/forms/post/post.ts
+++ b/src/app/components/forms/post/post.ts
@@ -392,7 +392,7 @@ export default class FormPost extends BaseForm<FormPostModel>
 			this.isSavedDraftPost = true;
 		}
 
-		// If the post is not a draft, default to posting to their profile.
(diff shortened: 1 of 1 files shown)


src/app/components/forms/post/post.ts
```diff
@@ -392,12 +392,15 @@ export default class FormPost extends BaseForm<FormPostModel>
 			this.isSavedDraftPost = true;
 		}
 
+		// If the post is not a draft, default to posting to their profile.
```
nilllzz, line 395: Yes, but why. Should be explaining like "Don't overwrite setting on a draft post because the user has already made a choice for this setting." or something like that.

nilllzz (approved):

src/app/components/forms/post/post.ts
hapurubokka: I like your wording more - updated

PR #604, opened 2020-08-04
Fix profile no-shouts illustration not appearing

hapurubokka:
Properly assigns a CommentStore lock when visiting profile overview route, allowing us to get the shout/comment count without having to open all shouts.


nilllzz:
Fixes https://github.com/gamejolt/issue-tracker/issues/1404

PR #605, opened 2020-08-06
Fix duplicate chat notifications

ttbowen:


src/app/components/chat/user-channel.ts
```diff
@@ -29,6 +35,12 @@ export class ChatUserChannel extends Channel {
 		this.on('friend_remove', this.onFriendRemove.bind(this));
 		this.on('notification', this.onNotification.bind(this));
 		this.on('you_updated', this.onYouUpdated.bind(this));
+		this.onClose(() => {
+			this.notificationChannel.close();
+			this.elector.die();
+		});
+
+		this.elector.awaitLeadership();
```
hworld, line 43: Do we need to actually await this somehow before further processing things? It returns a promise.

PR #607, opened 2020-08-11
Allow general media items to be used in lightbox

hapurubokka:


src/_common/content/components/media-item/media-item.ts
```diff
@@ -101,10 +109,6 @@ export default class AppContentMediaItem extends Vue {
 		return this.computedHeight > 0 ? this.computedHeight + 'px' : 'auto';
 	}
 
-	get isHydrated() {
```
nilllzz, line 104: Why should we remove this? I feel like expressing when something is loaded/hydrated through a prop makes more sense than checking the specific case of `!!this.mediaItem` everywhere.
It's also better to have a getter in case the check ever changes.

src/_common/content/content-viewer/content-viewer.ts
```diff
@@ -79,6 +94,43 @@ export default class AppContentViewer extends Vue implements ContentOwner {
 	setContent(content: ContentDocument) {
 		this.data = content;
 		this.hydrator = new ContentHydrator(content.hydration);
+		const processedItems: any[] = [];
+
+		this.data.content.forEach(item => {
+			if (item.type === 'mediaItem' && !item.attrs.href) {
+				processedItems.push(
+					new AppContentMediaItem({
```
nilllzz, line 102: Instead of this utter insanity, we should export the creation of the media item object from the `created` function of that vue component and just call that function here and in the create function.

src/_common/content/content-viewer/content-viewer.ts
```diff
@@ -79,6 +94,43 @@ export default class AppContentViewer extends Vue implements ContentOwner {
...
+			this.mediaItemsTemp = processedItems;
+		} else {
+			this.mediaItems = processedItems;
+			this.mediaItemsTemp = null;
+		}
+	}
+
+	setActiveItem(item: any) {
+		let index = item;
+		if (typeof item === 'object') {
```
nilllzz, line 129: Wouldn't the item need to match the signature of a `MediaItem`? Ie we should be able to check `instanceof MediaItem`, right?

nilllzz (changes requested):

src/_common/content/components/media-item/media-item.ts
hapurubokka: The only reason I did this is because the formatter was showing warnings in the Vue template that `mediaItem` could be `null` when checking for `isHydrated`. I agree that it previously looked better, but `isHydrated` wasn't seeming to pass on that information down the tree in the template.

hworld: I think we should keep it with the getter. The templates complaining is pretty annoying, but I feel like we need a different solution for it. We could always make a new getter that's like hydratedMediaItem that would return the mediaItem as non-null, and then using that in the template to remove the warnings, haha

src/_common/lightbox/item/item.styl
```diff
@@ -48,12 +49,21 @@ img
 	width: 100%
 	position: relative
 
+	// Make room for slider controls on Xs breakpoint - helps with issues like iframes
+	// breaking vue-touch and allows us to use arrows as an indication of more content.
+	@media $media-xs
```
hworld, line 54: I kind of think we should show the arrows at the bottom of the lightbox window on mobile if we are gonna show them, instead of left/right of the media item. The wider we can show an image, the better, I think. What are your thoughts on that?

src/_common/lightbox/item/item.ts
```diff
@@ -46,6 +46,14 @@ export default class AppLightboxItem extends Vue {
 		return this.isActive;
 	}
 
+	get isAnimatedGif() {
```
hworld, line 49: Maybe call this `isGifWithoutVideo` or something. I was confused at first

hworld (changes requested):

src/_common/lightbox/lightbox.styl
```diff
@@ -31,48 +30,69 @@ $-controls-padding = 10px
...
 	z-index: 1
 	background-color: rgba($black, 0.5)
-	text-align: center
+	display: flex
+	justify-content: center
+	align-items: center
 
 	@media $media-sm-up
-		display: block
+		rounded-corners-lg()
```
hworld, line 67: I think we should instead float the next/prev controls to the bottom on mobile so that it doesn't take up space from left/right. Maybe we should just disable the lightbox dragging on mobile if it's not really working well?

src/_common/content/components/media-item/media-item.ts
```diff
@@ -8,8 +9,21 @@ import { AppObserveDimensions } from '../../../observe-dimensions/observe-dimens
 import { AppTooltip } from '../../../tooltip/tooltip-directive';
 import { ContentEditorLinkModal } from '../../content-editor/modals/link/link-modal.service';
 import { ContentOwner } from '../../content-owner';
+import AppContentViewerTS from '../../content-viewer/content-viewer';
+import AppContentViewer from '../../content-viewer/content-viewer.vue';
 import AppBaseContentComponent from '../base/base-content-component.vue';
 
+export function createMediaItem(owner: ContentOwner, mediaItemId: number) {
```
hworld, line 16: name this longer like: `createMediaItemFromContentOwner`

hworld (changes requested):

src/_common/content/components/media-item/media-item.ts
```diff
@@ -142,21 +167,19 @@ export default class AppContentMediaItem extends Vue {
...
-		});
+		this.mediaItem = createMediaItem(this.owner, this.mediaItemId);
+		if (!this.mediaItem) {
+			this.hasError = true;
+		}
 	}
 
 	mounted() {
 		this.computeSize();
+		this.contentViewerParent = findVueParent(this, AppContentViewer) as AppContentViewerTS;
```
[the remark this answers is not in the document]
hapurubokka: I think we want this to be optional because when editing the media items, they have no contentViewer parent - when I was doing `findRequiredVueParent`, things were throwing errors and breaking because of this.

Changing to `findVueParent<AppContentViewerTS>(this, AppContentViewer);`

src/_common/lightbox/lightbox.styl
hapurubokka: I moved the controls to the bottom for phones, but didn't change the dragging functionality.

src/_common/content/components/media-item/media-item.ts
```diff
@@ -13,15 +13,19 @@ import AppContentViewerTS from '../../content-viewer/content-viewer';
...
-		if (data) {
-			item = new MediaItem(data);
-		}
-	});
-
-	return item;
+export async function createMediaItemFromContentOwner(owner: ContentOwner, mediaItemId: number) {
+	let item: MediaItem | null = null;
+
+	return owner
```
hworld, line 19: If useData is a promise, just await this.

```
return await owner.getHydrator().useData(...);
```

However, you're not really using the functionality of the async/await system, so just return without await or marking the function async. The return of the Promise will let you know it's async.

src/_common/content/components/media-item/media-item.ts
```diff
@@ -164,8 +168,9 @@ export default class AppContentMediaItem extends Vue {
 		return !this.mediaItem?.is_animated && !!this.mediaItem?.mediaserver_url;
 	}
 
-	created() {
-		this.mediaItem = createMediaItemFromContentOwner(this.owner, this.mediaItemId);
+	async created() {
+		this.mediaItem = await createMediaItemFromContentOwner(this.owner, this.mediaItemId);
+		console.log(12, this.mediaItem);
```
hworld, line 173: hello

src/_common/content/components/media-item/media-item.ts
```diff
@@ -164,8 +168,9 @@ export default class AppContentMediaItem extends Vue {
 		return !this.mediaItem?.is_animated && !!this.mediaItem?.mediaserver_url;
 	}
 
-	created() {
-		this.mediaItem = createMediaItemFromContentOwner(this.owner, this.mediaItemId);
+	async created() {
```
hworld, line 171: shouldn't make these hooks async, it also won't properly assign in SSR I think? does this make it so taht the media doesn't show anymore in SSR?

src/_common/content/components/media-item/media-item.ts
hapurubokka: I was having some issues making `useData(...)` work through a general function this way, so I had to revert it. That means there's slightly duplicate code, but seems to align more properly with how `owner.getHydrator().useData(...)` is intended to be used.

src/_common/content/content-viewer/content-viewer.ts
```diff
@@ -22,6 +27,18 @@ export default class AppContentViewer extends Vue implements ContentOwner {
 	data: ContentDocument | null = null;
 	hydrator: ContentHydrator = new ContentHydrator();
 
+	isInview = GJ_IS_SSR;
+	mediaItemsNeedRefreshed = true;
```
YLivay, line 31: This is a bit misleading. It's not that it's in view if we're doing SSR, its that if we're doing SSR we need to fetch the media items regardless of whether its in view. I think we should check for GJ_IS_SSR in setContent instead of this. feels hacky.

YLivay (changes requested):

src/_common/content/content-viewer/content-viewer.ts
```diff
@@ -79,6 +96,54 @@ export default class AppContentViewer extends Vue implements ContentOwner {
...
+			if (item.attrs.href) {
+				return;
+			}
+
+			let _mediaItem: MediaItem | null = null;
+			this.owner.getHydrator().useData('media-item-id', item.attrs.id.toString(), data => {
+				if (data) {
+					_mediaItem = new MediaItem(data);
+				}
+			});
```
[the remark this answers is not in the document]
hapurubokka: renaming to `processedItem`

src/_common/content/components/media-item/media-item.vue
```diff
@@ -29,7 +33,7 @@
 					:media-item="mediaItem"
 					radius="lg"
 				>
-					<template v-if="isHydrated">
+					<template v-if="mediaItem">
```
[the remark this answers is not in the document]
hapurubokka: I think @hworld and I thought it would be best to not have the getter, just so the template isn't giving us incorrect warnings. I did remove the `!!` though, as that was unneeded.

src/_common/content/components/media-item/media-item.ts
hapurubokka: This was breaking SSR - change to a different usage of creating the media item.

src/_common/content/content-viewer/content-viewer.ts
```diff
@@ -97,49 +98,53 @@ export default class AppContentViewer extends Vue implements ContentOwner, Light
...
 				if (data) {
-					_mediaItem = new MediaItem(data);
+					processedItem = new MediaItem(data);
 				}
 			});
 
-			if (_mediaItem) {
-				processedItems.push(_mediaItem);
+			if (processedItem) {
+				processedItems.push(processedItem);
```
YLivay: This way of retruning processedItems wont work. If hydrator doesn't have the data it'll add the items to the array only after the function returns it which is unexpected.

Since you need the getMediaItems to be syncronous for SSR I don't think you can promisify it..
Lets talk about this.

YLivay (changes requested):

hworld (approved):

PR #612, opened 2020-08-17
Chat edit improve

nilllzz:
- Puts cursor at the end of the doc when starting to edit
 - Pressing ESC in chat will refocus the editor after clearing message


src/_common/content/content-editor/content-editor.ts
```diff
@@ -352,7 +354,15 @@ export default class AppContentEditor extends Vue implements ContentOwner {
...
-			this.createView(state);
+
+			const view = this.createView(state);
+
+			// Set selection at the end of the document.
+			const tr = view.state.tr;
+			const selection = Selection.atEnd(view.state.doc);
+			tr.setSelection(selection);
+			tr.scrollIntoView();
+			view.dispatch(tr);
```
hapurubokka, line 365: This is currently only working for messages that don't contain uploaded media or tenor gifs - I'm assuming because the media is only taking up the equivalent of a paragraph when we're scrolling the selection into view. Setting the function as `async` and adding `await this.nextTick()` here seems to fix the issue though.
```suggestion
			tr.setSelection(selection);
			await this.nextTick();
			tr.scrollIntoView();
			view.dispatch(tr);
```

PR #623, opened 2020-08-28
Improve the cbar-menu and left-pane usage for phone sizes

hapurubokka:


src/_common/nav/navbar/navbar.styl
```diff
@@ -22,23 +22,24 @@ $navbar-height = $shell-top-nav-height is defined ? $shell-top-nav-height : 50px
...
 	overflow: hidden
 	height: $navbar-height
-	line-height: 20px
+	line-height: 26px
 	user-select: none
-	padding: 15px 10px
-
-	@media $media-sm-up
-		padding: 15px
+	padding: 15px // 10px
```
hworld, line 39: you left some comments here

src/app/store/index.ts
```diff
@@ -554,6 +557,10 @@ export class Store extends VuexStore<Store, Actions, Mutations> {
 			this.overlayedLeftPane = type;
 		}
 
+		if (type.length > 0) {
```
hworld, line 560: I think an empty string is falsy, yeah? so just check `if (type) {`

PR #625, opened 2020-08-29
Fix account-popover active indicator

hapurubokka:


src/app/components/shell/account-popover/account-popover.vue
```diff
@@ -1,27 +1,30 @@
...
 		popover-class="fill-darkest"
 		fixed
 		hide-on-state-change
 		@show="onShow()"
 		@hide="onHide()"
-		v-app-track-event="`top-nav:user-menu:toggle`"
 	>
 		<a class="navbar-item navbar-avatar" :class="{ active: isShowing }">
 			<app-user-avatar-img :user="app.user" />
 			<div
```
nilllzz, line 15: I am not 100% sure, but I don't think you can self-close div tags in HTML5.

src/app/components/shell/account-popover/account-popover.vue
```diff
@@ -30,50 +33,50 @@
...
 						<translate>Edit Account</translate>
 					</router-link>
 					<router-link
+						v-app-track-event="`account-popover:stickers`"
 						class="list-group-item offline-disable"
 						:to="{ name: 'dash.stickers.overview' }"
-						v-app-track-event="`account-popover:stickers`"
 					>
 						<translate>Stickers</translate>
-						<span v-if="shouldShowNewStickers" class="-new-tag -new-tag-list"></span>
```
nilllzz, line 58: Also can't self-close spans.
Reference: https://html.spec.whatwg.org/multipage/syntax.html#start-tags (specifically 6.)
Valid self-closing elements are the "void elements" in this specification: https://html.spec.whatwg.org/multipage/syntax.html#void-elements

nilllzz (changes requested):

hworld (approved):

PR #626, opened 2020-08-31
Generate one system notification for chat messages

ttbowen:
Fix so all tabs show a notification in-page, but only spawn one system notification.


hworld:
Incorporated into this PR https://github.com/gamejolt/gamejolt/pull/627

PR #631, opened 2020-09-07
Add an unread blip to inline-community-card

hapurubokka:
image
image

### DIFF
--- a/src/_common/community/card-base/card-base-inline.styl
+++ b/src/_common/community/card-base/card-base-inline.styl
@@ -5,6 +5,20 @@ $-thumbnail-size = 80px
 $-thumbnail-size-sm = 48px
 $-bg-color-base = var(--theme-bg)
 
+.-unread-blip
+	position: absolute
+	top: 16px
+	left: 24px
+	height: 12px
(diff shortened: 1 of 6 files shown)


src/_common/community/card-base/card-base-inline.styl
```diff
@@ -5,6 +5,20 @@ $-thumbnail-size = 80px
...
 $-bg-color-base = var(--theme-bg)
 
+.-unread-blip
+	position: absolute
+	top: 16px
+	left: 24px
+	height: 12px
+	width: 12px
+	border: $border-width-large solid var(--theme-bg)
+	background-color: var(--theme-fg)
```
YLivay, line 15: I think we should use theme-notice for this one. Thats the color of the blips on the channel images as well.

YLivay (approved):

src/_common/community/card-base/card-base-inline.styl
hworld: I decided against that since it's not an actual "notification" it's just "unread"

hworld (approved):

PR #639, opened 2020-09-16
Lazy load post articles.

hworld:


src/app/components/activity/feed/devlog-post/text/text.ts
```diff
@@ -13,22 +17,22 @@ import { ActivityFeedView } from '../../view';
...
-	@Prop(ActivityFeedItem)
-	item!: ActivityFeedItem;
+	@Prop(ActivityFeedItem) item!: ActivityFeedItem;
+	@Prop(FiresidePost) post!: FiresidePost;
 
-	@Prop(FiresidePost)
-	post!: FiresidePost;
-
-	isToggling = false;
+	isLoaded = !!this.post.article_content;
```
YLivay, line 29: rename to more specific `isArticleLoaded`. Also, we can set it to true even if article_content isn't populated if we check against has_article, yeah?

src/app/components/activity/feed/devlog-post/text/text.ts
```diff
@@ -76,9 +79,5 @@ export default class AppActivityFeedDevlogPostText extends Vue {
 		}
 
 		this.feed.setItemOpen(this.item, false);
-
-		setTimeout(() => {
-			this.isToggling = false;
-		}, 1000);
```
YLivay, line 82: Was that meant purely to give the animation time to complete or something? why was that even there?

src/app/components/activity/feed/devlog-post/text/text.ts
```diff
@@ -59,9 +58,13 @@ export default class AppActivityFeedDevlogPostText extends Vue {
 		}
 	}
 
-	expand() {
+	async expand() {
 		this.feed.setItemOpen(this.item, true);
```
YLivay, line 62: Don't we want to the set feed item as open only after the article was loaded in?

src/app/components/forms/post/post.vue
```diff
@@ -292,14 +288,14 @@
 				</div>
 
 				<p v-if="pollDuration < MIN_POLL_DURATION" class="help-block error">
-					<translate
-						>Too short! Polls must be between 5 minutes and 14 days long.</translate
-					>
```
YLivay, line 297: TODO: test that message isnt wonky due to whitespace being preserved

src/app/components/forms/post/post.vue
```diff
@@ -532,11 +527,11 @@
 
 					<template v-if="!wasPublished && canAddCommunity">
 						<app-form-post-community-pill-add
-							class="-community-pill anim-fade-in-enlarge no-animate-leave"
 							key="add"
+							v-app-scroll-when="scrollingKey"
```
YLivay, line 531: well, if you're already at it, move this attribute above the `key` attribute.

src/app/components/forms/post/post.ts
```diff
@@ -440,14 +425,16 @@ export default class FormPost extends BaseForm<FormPostModel>
...
+		if (model.has_article) {
 			this.longEnabled = true;
 		}
 
 		await this.fetchTimezones();
 	}
 
 	onLoad(payload: any) {
+		// Pull any post information that may not already be loaded in.
+		this.formModel.assign(payload.post);
```
YLivay, line 437: This breaks creating posts since it sets the status to temp, which gets sent back during save.
What data exactly do we want to fetch here?

YLivay: specifically populating article_content works.

YLivay (changes requested):

src/_common/fireside/post/post-model.ts
```diff
@@ -511,3 +520,12 @@ export class FiresidePost extends Model implements ContentContainerModel, Commen
...
 Model.create(FiresidePost);
+
+/**
+ * Will load the article from the API and store it into the post model.
+ */
+export async function loadArticleIntoPost(post: FiresidePost) {
+	const payload = await Api.sendRequest(`/web/posts/article/${post.id}`);
+	post.article_content = payload.article;
+	return post;
+}
```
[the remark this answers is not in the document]
hworld: Remember, we want to switch to more "functional" style so that things can be imported into the chunks they're used. When it's all instance functions that code gets imported always.

src/app/components/activity/feed/devlog-post/text/text.ts
hworld: This component is literally just a component to show the article. I would normally agree, but since it's scoped to just working with the article, I don't think we need to name it longer.

Also, has_article doesn't mean that article content is loaded in. It means that the article in the DB has content. I should probably change this check to use `post.hasArticleContent` which works off the loaded content to check.

src/app/components/activity/feed/devlog-post/text/text.ts
hworld: This way old code that I noticed we didn't need anymore since it had no effect. It used to animate the expanding and collapsing, but now it's instant.

src/app/components/activity/feed/devlog-post/text/text.ts
hworld: Nah, we need to open it up to show the loading graphic. If it's not set to "open" it won't show that it's loading. I guess we can change it so that it'll open up if it's loading OR if it's open? It would just complicate the checks a bit more, so I kept it simple like this.

I suppose there could be a race condition that they try opening the article and then move away from the page and the article never gets content somehow...

src/app/components/forms/post/post.vue
hworld: that's not the styleguide. key is always before since it's very important vue-wise. all of these little changes get auto-fixed with one command, btw.

src/app/components/forms/post/post.ts
hworld: Dang. It's mainly for article_content, but I didn't want it to be just for that. I wanted it generic to essentially make sure the post is fully loaded from server before you can work on it. Just in case there's more lazy loaded fields.

PR #641, opened 2020-09-17
Make the open article flow feel nicer by moving the height around less.

hworld:


src/app/components/activity/feed/devlog-post/text/text.vue
```diff
@@ -8,12 +8,14 @@
 		}"
 	>
 		<template v-if="isOpen">
-			<div class="page-cut" />
+			<div class="page-cut">
+				<div class="page-cut-content-placeholder" />
```
hworld, line 12: won't this show an overlay on top of the border as a cutout? why do we need this?

src/app/components/activity/feed/devlog-post/text/text.vue
```diff
@@ -8,12 +8,14 @@
...
+			</div>
 
 			<app-content-viewer :source="post.article_content" />
 		</template>
 
-		<div class="page-cut">
+		<div class="page-cut page-cut-bottom">
```
hworld, line 18: hyphenate `-page-cut-bottom` so that it doesn't conflict with page-cut in future

src/app/components/activity/feed/devlog-post/text/text.vue
hworld: also, hyphenate

hapurubokka: The placeholder is just a transparent div with the same height as the `page-cut-content` class. This aligns the top page-cut when the article expands and the article expansion control button is moved to the bottom.

hapurubokka: I mean that it prevents shifting of the page-cut when the article expands. Otherwise the page-cut would shift up when it loses the button child.

PR #645, opened 2020-09-20
Add 'Korean' Language

alexkoala:
This PR adds Korean(한국어) in language picker.


alexkoala:
L10N system has broken so it needs to be fixed first.

PR #647, opened 2020-09-21
Fix leaving group rooms

ttbowen:
Now handling the `group_leave` on user channel. This is so we can leave the group without joining the room channel we're leaving.

### DIFF
--- a/src/app/components/chat/user-list/item/item.ts
+++ b/src/app/components/chat/user-list/item/item.ts
@@ -142,6 +142,6 @@ export default class AppChatUserListItem extends Vue {
 			return;
 		}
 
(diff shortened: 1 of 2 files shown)


src/app/components/chat/user-list/item/item.ts
```diff
@@ -142,6 +142,6 @@ export default class AppChatUserListItem extends Vue {
 			return;
 		}
 
-		leaveGroupRoom(this.chat);
+		leaveGroupRoom(this.chat, this.item);
```
hworld, line 145: TS should've thrown an error here since item is either a chat room or a chat user.

hworld (approved):

PR #650, opened 2020-09-30
Grid notification update

nilllzz:


src/app/views/communities/view/view.ts
```diff
@@ -166,26 +168,26 @@ export default class RouteCommunitiesView extends BaseRouteComponent {
...
+
+		if (community.is_member) {
+			this.getCommunityBootstrap();
+		}
+	}
+
+	private async getCommunityBootstrap() {
+		// When this is the first route the user enters, grid might not be bootstrapped yet.
+		while (!this.grid) {
+			await sleep(250);
```
hworld, line 188: just need a grid promise we can await

src/app/components/activity/feed/feed.vue
```diff
@@ -1,15 +1,21 @@
...
 		Basically anytime the feed state's items are replaced so that the references
 		to them get picked up again.
 	-->
-	<div class="activity-feed" :key="feed.id">
+	<div :key="feed.id" class="activity-feed">
 		<template v-if="newCount > 0 || feed.isLoadingNew">
 			<app-scroll-inview :margin="`-${Scroll.offsetTop}px`" @inview="onNewButtonInview">
 				<app-expand v-if="!feed.isLoadingNew" :when="isNewButtonInview">
 					<app-activity-feed-new-button @click="loadNew()">
+						<translate v-if="!feed.showNewCountNumber">
```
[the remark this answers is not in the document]
hworld: After looking at all the code, I really do think we should simplify this state to show "load new" and never show or store the count. It keeps one flow in the code and reduces the amount of bookkeeping we need to care about.

hworld (changes requested):

src/app/components/grid/client.service.ts
```diff
@@ -75,14 +75,81 @@ type ClearNotificationsType =
...
+
+/**
+ * List of resolvers waiting for grid to connect.
+ * These resolvers get resolved in the connect function once Grid connected.
+ */
+let connectionResolvers: (() => void)[] = [];
+/**
+ * Resolves once Grid is fully connected.
+ */
+function tillConnection(client: GridClient) {
```
hworld, line 143: Can't we use just one promise instead of an array? We're resolving all of them with the same exact value essentially in this function and the grid bootstrap resolver

hworld (changes requested):

src/app/components/grid/client.service.ts
nilllzz: I think it's more easily understandable to keep a collection of resolvers and then resolve them all then to respond with one object for everybody to wait on and then resolve it. 
I don't really think it makes a huge difference in terms of usability or performance, so I'd go with this option.

PR #655, opened 2020-10-14
Chat invites

ttbowen:


src/app/components/chat/message.ts
```diff
@@ -1,13 +1,14 @@
 import { ChatUser } from './user';
 
-export type ChatMessageType = 0 | 1;
+export type ChatMessageType = 'normal' | 'system' | 'invite';
```
nilllzz, line 3: We don't import this anywhere, no export needed.

src/app/components/chat/message.ts
```diff
@@ -1,13 +1,14 @@
...
 
 export const CHAT_MESSAGE_MAX_CONTENT_LENGTH = 1000;
 export const TIMEOUT_CONSIDER_QUEUED = 1500; // Time in ms until a queued message should be displayed as such.
 
 export class ChatMessage {
-	static readonly TypeNormal = 0;
-	static readonly TypeSystem = 1;
+	static readonly TypeNormal = 'normal';
+	static readonly TypeSystem = 'system';
+	static readonly TypeInvite = 'invite';
```
nilllzz, line 11: I think what we should do here is use a const enum instead of these declarations:

Declare the const enum above the ChatMessage class like this:
```
export const enum ChatMessageType {
  NORMAL = 'normal',
  INVITE = 'invite',
}
```

In the ChatMessage class, the type definition for `type` stays the same:
```
type!: ChatMessageType;
```

Then, you can remove the `static readonly Type*` fields.

In ChatClient (line 426), you can then use:
```
type: ChatMessageType.NORMAL,
```
instead of 
```
type: ChatMessage.TypeNormal,
```

And for when you want to use the const enum in Vue view components:
In the TS file, import the const enum type into the local context:

```
readonly ChatMessageType = ChatMessageType;
```

And then you can use it like that in a Vue file:
```
		:class="{
			'chat-msg-type-normal': message.type === ChatMessageType.NORMAL,
```

In addition to all that, I don't think type "system" is in use, so you can remove it.

src/app/components/chat/client.ts
```diff
@@ -694,8 +692,18 @@ export function editMessage(chat: ChatClient, message: ChatMessage) {
...
+	if (room) {
+		chat.roomChannels[room.id]
+			.push('accept_invite', { msg_id: msgId })
+			.receive('ok', response => enterChatRoom(chat, response.room.id));
+	}
+}
+
 export function startTyping(chat: ChatClient) {
 	const room = chat.room;
+	F;
```
nilllzz, line 706: F;

src/app/components/chat/room.ts
```diff
@@ -41,9 +40,6 @@ export class ChatRoom {
 	}
 }
 
-export function getChatRoomTitle(room: ChatRoom, chat: ChatClient) {
-	return room.members
-		.filter(member => member.id !== chat.currentUser?.id)
-		.map(member => member.display_name)
-		.join(', ');
+export function getChatRoomTitle(room: ChatRoom) {
+	return room.members.map(member => member.display_name).join(', ');
```
nilllzz, line 44: Why is this changed? I'd think we want to not show the logged in user name in room titles.

src/app/components/chat/window/output/item/item.vue
```diff
@@ -6,6 +6,7 @@
 		:class="{
 			'chat-msg-type-normal': message.type === ChatMessage.TypeNormal,
 			'chat-msg-type-system': message.type === ChatMessage.TypeSystem,
```
nilllzz, line 8: Here too, remove type system, not used.
Make sure to also remove the `chat-msg-type-system` declaration in the CSS file.

src/app/components/chat/window/output/item/item.vue
```diff
@@ -103,7 +108,26 @@
...
+				</template>
+				<template v-else-if="message.type === ChatMessage.TypeInvite">
+					<div class="alert">
+						<p>
+							<translate>
+								You have been invited to a group.
+							</translate>
+						</p>
+						<app-button
+							:disabled="chat.currentUser && message.user.id === chat.currentUser.id"
```
nilllzz, line 122: I think instead of just disabling the accept button for the sending user, we should show a completely different message to the sender:
Have the text say something like "You have invite this user to a group".
I am not sure if we want to support canceling an invite, but that button would also go here I guess.

nilllzz: Also, make that check a getter on the component class, something like `isInviteSender`.

nilllzz (changes requested):

src/app/components/chat/room.ts
ttbowen: Because it means the room title is blank when the room is initially created.

PR #657, opened 2020-10-15
Split test to remove inline comments in feed, better tracking of event labels, don't show user follow widget in feed anymore.

hworld:


src/app/components/split-test/split-test-service.ts
```diff
@@ -1,20 +1,34 @@
...
+
+// 	if (typeof payload._experiment !== 'undefined' && typeof payload._variation !== 'undefined') {
+// 		if (payload._experiment === experiment) {
+// 			return payload._variation;
+// 		}
+// 	}
+
+// 	return -1;
+// }
+
```
hapurubokka, line 31: are we keeping this for later?

hworld: yes

hapurubokka (approved):

PR #661, opened 2020-10-20
Feed per page

nilllzz:
Allow control of feed per-page information through backend payloads.


src/app/components/activity/feed/view.ts
```diff
@@ -277,7 +295,7 @@ export class ActivityFeedView {
 		Analytics.trackEvent('activity-feed', 'loaded-more', 'page-' + this.totalTimesLoaded);
 	}
 
-	async loadNew(newCount = ItemsPerPage) {
+	async loadNew(newCount = this.itemsPerPage) {
```
YLivay, line 298: matter of preference but i dont like using these. can we just default to null and use a null coalesce in the method body?

src/app/components/shell/notification-popover/notification-popover.ts
```diff
@@ -64,6 +64,8 @@ export default class AppShellNotificationPopover extends Vue {
 				slice: 15,
 				shouldScroll: false,
 				shouldShowUserCards: false,
+				// Some default value before the real feed data gets populated on show.
+				itemsPerPage: 15,
```
YLivay, line 68: why do we need this default?
I mean, what happens when the feed attempts to load with an unspecified amount of items per page?

src/app/components/activity/feed/view.ts
nilllzz: Ehh, would you do something like
```
newCount = newCount ?? this.itemsPerPage;
```
?

src/app/components/shell/notification-popover/notification-popover.ts
nilllzz: I built a check that this parameter is definitely set into the feed view, to make sure we don't fuck it up and forget.
This is the only place in the code where the feed gets bootstrapped without a feed payload, so in this specific instance we just set a count here.
It does get populated later, as the comment says, so it doesn't really matter.
The `15` is never actually used because when the feed gets populated, it also receives the perPage info from backend at the same time.

src/app/components/activity/feed/view.ts
YLivay: yeah

YLivay (approved):

nilllzz:
Merged with liked posts feed

PR #666, opened 2020-10-22
Soft fail eject post

nilllzz:
When using communities with more than one client (tabs, applications, devices, etc), ejecting the same post "twice" results in the second attempt redirecting to an error page.
This PR changes that behavior so it instead fails silently and ejects the post in frontend anyway.


hworld:
I think this should be handled by backend instead, since we'll have the same issue in the app and it would be nice to handle it in one place (the backend).

I looked at the reject code. Essentially, if there is no fireside post community I think we can return success instead of the NOT_FOUND ApiException

nilllzz:
Handling in backend instead.

nilllzz:
New PR: https://github.com/gamejolt/gamejolt/pull/705

PR #667, opened 2020-10-23
Room titles

ttbowen:


src/app/components/chat/invite-modal/invite-modal.vue
```diff
@@ -16,6 +16,7 @@
 			</div>
 
 			<div class="modal-body">
+				<input class="-filter form-control" placeholder="Title" />
```
hapurubokka, line 19: This doesn't appear to be tied to anything. Trying to initialize a group chat with a room title doesn't name it.

hapurubokka: This component will probably need to be changed to an `app-form` input field.

src/app/components/chat/room-details-modal/form/form.ts
```diff
@@ -0,0 +1,20 @@
...
+import { BaseForm } from '../../../../../_common/form-vue/form.service';
+import { ChatRoom } from '../../room';
+
+export type FormModel = {
+	title: string;
+};
+
+@Component({})
+export default class FormRoomDetails extends BaseForm<FormModel> {
+	modelClass = ChatRoom;
```
hapurubokka, line 11: This doesn't appear to be used for anything.

src/app/components/chat/room-details-modal/room-details-modal.vue
```diff
@@ -1,3 +1,4 @@
+<script lang="ts" src="./room-details-modal"></script>
 <template>
```
hapurubokka, line 2: add a space between these

src/app/components/chat/room-details-modal/form/form.ts
```diff
@@ -0,0 +1,20 @@
...
+
+	@Emit('submit')
+	emitSubmit(_title: FormModel) {}
+
+	onSubmit() {
+		const submit: FormModel = { title: this.formModel.title };
```
hapurubokka, line 17: Typescript should already have the typing information - we shouldn't need to specifically describe a type here.

src/app/components/chat/room-details-modal/form/form.ts
```diff
@@ -0,0 +1,20 @@
+import { Component, Emit } from 'vue-property-decorator';
+import { BaseForm } from '../../../../../_common/form-vue/form.service';
+import { ChatRoom } from '../../room';
+
+export type FormModel = {
```
hapurubokka, line 5: If we're exporting this, I think we should rename it to relate to the scope it represents so it doesn't get confusing when importing.

src/app/components/chat/room-details-modal/form/form.ts
hapurubokka: actually, I think we should just emit the entire `formModel` object instead of extracting the title like this. The function that triggers from the emit can then extract any information it needs from the event and do what it needs to with it.

hapurubokka (changes requested):

PR #669, opened 2020-10-26
Add spacing for video play counts for small screens

hapurubokka:


src/app/views/post/_page/page.vue
```diff
@@ -248,6 +248,9 @@
 		display: inline-flex
 		align-items: center
 
+		@media $media-xs
+			margin-right: 8px
```
hworld, line 252: should be gutter-width-xs / 2

hworld (approved):

PR #679, opened 2020-10-27
Better mute handling in video-player volume controls

hapurubokka:


src/_common/video/player/player.vue
```diff
@@ -30,7 +30,7 @@
 
 		<transition>
 			<div
-				v-if="shouldShowUI"
+				v-if="shouldShowUI || true"
```
hworld, line 33: nani?!

src/_common/video/player/slider/slider.ts
```diff
@@ -27,6 +27,14 @@ export default class AppVideoPlayerSlider extends Vue {
 		thumb: HTMLDivElement;
 	};
 
+	get thumbOffset() {
```
hworld, line 30: I think a lot of code in this file would simplify if player was a required prop.

src/_common/video/player/controller.ts
```diff
@@ -53,31 +62,56 @@ export function toggleVideoPlayback(player: VideoPlayerController) {
 	}
 }
 
-export function setVideoVolume(player: VideoPlayerController, level: number) {
+export function setVideoVolume(
```
hworld, line 65: I think we should keep this function very simple as just a way to set the direct video volume. I think we should have a new function that is maybe scrubVideoVolume that works similar to how we scrub the current time. That can do the proper setting of fallback volume and toggling of the video muted state. It would be a function that calls into this function but also does extra work depending on the scrubbing state.

src/_common/video/player/controller.ts
```diff
@@ -53,31 +62,56 @@ export function toggleVideoPlayback(player: VideoPlayerController) {
...
+export function toggleVideoMuted(player: VideoPlayerController) {
+	player.isMuted = !player.isMuted;
+	if (!player.isMuted && player.mutedFallbackVolume !== null) {
+		setVideoVolume(player, player.mutedFallbackVolume, false);
+	}
+
+	if (player.context === 'feed') {
+		SettingVideoPlayerFeedMuted.set(player.isMuted);
+		// Feed videos don't have a volume slider, so if the video element volume was somehow changed we want to set it back to 100%.
+		if (!player.isMuted) {
```
hworld, line 102: I think you should move this condition back up into the first if of the function:
```
if (!player.isMuted) {
    if (player.mutedFallbackVolume) {
        setVideoVolume(player, player.mutedFallbackVolume);
    } else if (player.context === 'feed') {
        setVideoVolume(player, 100);
    }
}
```

hworld (changes requested):

PR #681, opened 2020-10-29
Fix user/game header forms allowing setCrop on removed header

nilllzz:
Both game and user forms had errors that allowed a user to set a crop on a header image that was already removed.


src/app/components/forms/game/header/header.ts
```diff
@@ -75,7 +75,11 @@ export default class FormGameHeader extends BaseForm<FormModel>
 		);
 
 		if (result) {
-			this.formModel.$clearHeader();
+			const payload = await this.formModel.$clearHeader();
+			// Overwrite the base model's header media item here.
+			// This needs to be done because this form does not resolve (and may never resolve)
+			// after clearning a header. Need to ensure that the base model's header gets cleared.
+			(this.model as any).header_media_item = payload.game.header_media_item;
```
hworld, line 82: I think we should just assign: `this.model?.assign(payload.game)`; That'll effectively refresh the game with the current data, which is what happens anyway during $clearHeader on the formModel.

hworld (changes requested):

src/app/components/forms/game/header/header.ts
```diff
@@ -75,7 +75,11 @@ export default class FormGameHeader extends BaseForm<FormModel>
 		);
 
 		if (result) {
-			this.formModel.$clearHeader();
+			const payload = await this.formModel.$clearHeader();
+			// Overwrite the base model's header media item here.
+			// This needs to be done because this form does not resolve (and may never resolve)
+			// after clearning a header. Need to ensure that the base model's header gets cleared.
```
YLivay, line 81: clearning :sponge:

PR #684, opened 2020-10-30
Sticker drawer bug-fixes and improvements

hapurubokka:


src/_common/sticker/target/target-controller.ts
```diff
@@ -26,6 +26,19 @@ export class StickerTargetController {
 
 	private _shouldShow = false;
 
+	get canPlaceStickers(): boolean {
```
hworld, line 29: Let's use the disabled prop for this.

We should make functions in StickerPlacement that can check `canPlaceOnPost` and `canPlaceOnComment`. The comment one should be able to take a parentComment too and should literally just call into `canCommentOnModel` for now. The post one should be this code.

You should use the same functions for checking if we should show the sticker button control as well for both comments and posts.

src/_common/sticker/target/target.ts
```diff
@@ -42,6 +42,10 @@ export default class AppStickerTarget extends Vue {
 	@Inject(DrawerStoreKey) drawerStore!: DrawerStore;
 	@Inject(StickerLayerKey) layer!: StickerLayerController;
 
+	get canPlaceStickers() {
```
hworld, line 45: revert this back to just disabled.

src/app/views/dashboard/stickers/stickers.ts
```diff
@@ -35,6 +35,8 @@ export type StickerCount = {
 	sticker: Sticker;
 };
 
+const FETCH_STICKERS = '/web/stickers/dash';
```
hworld, line 38: name it FetchStickersEndpoint. I try not to ALLCAPS in TS, but instead use TitleCase

src/app/views/dashboard/stickers/stickers.ts
```diff
@@ -99,14 +101,26 @@ export default class RouteDashStickers extends BaseRouteComponent {
 	}
 
 	routeResolved($payload: InitPayload) {
-		this.balance = $payload.balance;
-		this.stickerCost = $payload.stickerCost;
+		this.fetchStickers($payload);
```
hworld, line 104: instead of doing all of this, you can actually just call `this.reloadRoute()`. It'll essentially rerun the route resolver, recall routeResolved, and push the new payload data into the cache for the route. =]

src/app/views/post/_page/page.ts
```diff
@@ -240,4 +240,11 @@ export default class AppPostPage extends Vue implements LightboxMediaSource {
 		this.lightbox.close();
 		this.lightbox = undefined;
 	}
+
+	@Watch('post.id')
+	onPostIdChange() {
+		if (this.stickerTargetController.model.id !== this.post.id) {
```
hworld, line 246: I think it's fine to always just refresh, even if the model somehow didn't change, right? Or is this if condition there for a reason?

src/app/views/post/_page/page.vue
```diff
@@ -102,7 +102,7 @@
 						-->
 						<div ref="sticker-scroll" />
 
-						<div v-if="post.hasMedia" class="-media-items">
+						<div v-if="post.hasMedia" :key="`media-${post.id}`" class="-media-items">
```
hworld, line 105: I think you should put a comment for this key and the other to say why we need it.

hworld (changes requested):

src/app/views/post/_page/page.ts
hapurubokka: It probably makes more sense to do this on any change with the post id - revised

hworld (approved):

PR #687, opened 2020-11-03
Add black-bar margins to narrow post-page videos

hapurubokka:


hworld (changes requested):
Video player stats is no longer styled properly.
Opening the video player up in the post edit modal breaks.
Video stats showing should be a prop that's false by default.

src/_common/video/player/player.ts
```diff
@@ -50,24 +60,56 @@ function createReadableTimestamp(time: number) {
...
 	@Prop(propRequired(Array)) manifests!: string[];
 	@Prop(propOptional(Boolean, false)) autoplay!: boolean;
 	@Prop(propOptional(Number, 0)) startTime!: number;
 	@Prop(propOptional(String, null)) context!: VideoPlayerControllerContext;
+	@Prop(propOptional(Number, 0)) viewCount!: number;
+	@Prop(propOptional(Boolean, false)) showVideoStats!: boolean;
 
-	player = new VideoPlayerController(this.poster, this.manifests, this.context);
-	isHovered = false;
+	player = new VideoPlayerController(undefined, this.manifests, this.context);
```
hworld, line 80: remove the poster field since I think we always manually put it behind the video now

hworld (changes requested):
Fullscreen

PR #691, opened 2020-11-06
Remove Shaka integration from feed video players

hapurubokka:
Removes Shaka from feed, uses the `AppVideo` component instead


src/_common/fireside/post/video/video-model.ts
```diff
@@ -46,6 +47,19 @@ export class FiresidePostVideo extends Model {
 			.sort((a, b) => getManifestPreference(a) - getManifestPreference(b))
 			.map(i => i.img_url);
 	}
+
+	get transcodedVideos(): VideoSourceArray {
+		const videoSources: VideoSourceArray = [];
+		this.media.map(i => {
```
hworld, line 53: change to `.filter.map` then you can return without having the intermediate array.

src/_common/game/thumbnail-img/thumbnail-img.ts
```diff
@@ -39,6 +40,23 @@ export default class AppGameThumbnailImg extends Vue {
 		return this.hasVideo && ContentFocus.hasFocus;
 	}
 
+	get videoSources(): VideoSourceArray {
```
hworld, line 43: needs to handle the case where mp4 url or webm url aren't set. they might be empty/null and we shouldn't return it as part of the array then

src/_common/lightbox/item/item.ts
```diff
@@ -62,6 +63,19 @@ export default class AppLightboxItem extends Vue {
 		return this.item.getMediaItem()!;
 	}
 
+	get videoSources(): VideoSourceArray {
```
hworld, line 66: needs to handle empty mp4/webm

src/_common/media-item/post/post.ts
```diff
@@ -61,6 +62,19 @@ export default class AppMediaItemPost extends Vue {
 		return this.isActive && ContentFocus.hasFocus;
 	}
 
+	get videoSources(): VideoSourceArray {
```
hworld, line 65: needs to handle empty mp4/webm

src/app/components/broadcast-modal/broadcast-modal.ts
```diff
@@ -41,4 +43,17 @@ export default class AppBroadcastModal extends BaseModal {
 	$refs!: {
 		modal: AppModalTS;
 	};
+
+	getVideoSources(item: MediaItem): VideoSourceArray {
```
hworld, line 47: needs to handle empty mp4/webm

src/_common/video/player/shaka.ts
```diff
@@ -29,42 +34,33 @@ export default class AppVideoPlayerShaka extends Vue {
...
-		this.trackPlaytime();
 		this.isDestroyed = true;
 		if (this.shakaPlayer) {
 			this.shakaPlayer.destroy();
 			this.shakaPlayer = undefined;
 		}
 	}
 
-	private async init() {
+	async initShakaWithVideo(video: HTMLVideoElement, syncFunction: () => void) {
```
hworld, line 53: I think instead of syncFunction, the AppVideo should expect this function to return either true or false (Promise<boolean>). This way it would know if the initialization failed or not. Just await it in AppVideo. No need to pass a sync function that we need to remember to call, I think.

src/_common/video/video.ts
```diff
@@ -81,4 +124,137 @@ export default class AppVideo extends Vue {
...
+	}
+
+	@Watch('shouldPlay')
+	onShouldPlayChange() {
+		// Ignore 'shouldPlay' changes for instances that have an inherited player controller handling playback.
+		if (this.inheritedPlayer) {
+			return;
+		}
+
+		if (this.shouldPlay) {
```
hworld, line 214: shouldn't we call the toggle playback function instead of setting it directly?

src/_common/video/video.ts
```diff
@@ -81,4 +124,137 @@ export default class AppVideo extends Vue {
...
+		// pause the video in the player controller and reset the player
+		// volume to the initial setting.
+		this.player.state = 'paused';
+		this.player.volume = startVolume;
+	}
+
+	@Watch('shouldPlay')
+	onShouldPlayChange() {
+		// Ignore 'shouldPlay' changes for instances that have an inherited player controller handling playback.
+		if (this.inheritedPlayer) {
```
hworld, line 210: I don't think we need this check. shouldPlay will never change anyway unless they bind to it

src/_common/video/video.ts
```diff
@@ -1,77 +1,120 @@
...
 	@Prop(propOptional(Boolean, true)) shouldPlay!: boolean;
+	@Prop(propOptional(Boolean, false)) shouldMute!: boolean;
+	@Prop(propOptional(Function)) initCallback?: (
+		videoTag: HTMLVideoElement,
+		syncFunction: () => void
+	) => void;
+
+	@Prop(propOptional(String, null)) playerContext!: VideoPlayerControllerContext;
+	@Prop(propOptional(VideoPlayerController, null))
+	inheritedPlayer!: VideoPlayerController | null;
```
hworld, line 39: I think you should instead just have this be `player` with a default value of a new video player controller. Instead of using this prop to figure out if we should do certain things, I think they should be split out into separate props to control the enlabling/disabling of the functionality.

src/_common/video/video.ts
```diff
@@ -1,77 +1,120 @@
...
+			// can clear it after.
+			const onPlay = () => {
+				this.isLoaded = true;
+				this.video.removeEventListener('play', onPlay);
+			};
+			this.video.addEventListener('play', onPlay);
+		}
+
+		this.syncVolume();
+		this.syncTime();
```
hworld, line 91: I'm not sure how well syncTime works before it has shaka loaded in, do you?

src/_common/video/video.ts
```diff
@@ -1,77 +1,120 @@
...
 
 		// As soon as we append, it'll actually load into view and start
 		// playing.
 		this.$el.appendChild(this.video);
+
+		if (this.initCallback) {
+			// If using a callback function (used to set up Shaka events with the right timing),
+			// we need to run 'this.syncStates' after Shaka is done initializing itself.
+			this.initCallback(this.video, this.syncStates);
```
hworld, line 100: yeah, await here before setting isLoaded = true, yeah?

src/_common/video/video.ts
```diff
@@ -81,4 +124,137 @@ export default class AppVideo extends Vue {
 			this.video.load();
 		}
 	}
+
+	private trackPlaytime() {
+		// Only track playtime events for components that get a VideoPlayerController passed in (meaning that we have playback controls).
+		if (!this.videoStartTime || !this.inheritedPlayer) {
```
hworld, line 130: control this functionality through a prop if we don't want to always track

hworld (changes requested):

src/_common/video/video.ts
hapurubokka: removed

hworld (approved):

PR #694, opened 2020-11-13
Fix message combine when removing a message

ttbowen:


src/app/components/chat/room-channel.ts
```diff
@@ -122,6 +122,10 @@ export class ChatRoomChannel extends Channel {
 	private onRemoveMsg(data: { id: number }) {
 		if (this.room) {
 			arrayRemove(this.client.messages[this.roomId], i => i.id === data.id);
+
+			const messages = [...this.client.messages[this.roomId]];
+			this.client.messages[this.roomId] = [];
+			processNewChatOutput(this.client, this.roomId, messages, false);
```
YLivay, line 128: Nope. we're not going to reprocess all loaded messages every time a message is removed + doesn't this actually spawn notifications if isHistorical arg is false here?

This needs to be done differently, going backwards from the removed messages to see if a new gap needs to be introduced and let vue reactivity do its thing.

YLivay (changes requested):

nilllzz:
Fixed in: https://github.com/gamejolt/gamejolt/pull/720

PR #695, opened 2020-11-17
Featured on activity

nilllzz:
Show the last featured game on the activity feed.


YLivay (changes requested):
Lets join these two requests for recommended users and games. This is a mighty hot code path.
If you absolutely want to keep the components separate we can talk about it.

YLivay (approved):

PR #716, opened 2020-12-03
Eject move notifications

nilllzz:
- Supports new event notification type to show the user community notifications
 - Shows notifications users receive on their posts through community mod actions


src/_common/user/action-reasons.ts
```diff
@@ -0,0 +1,71 @@
...
+ */
+function getReasons(reasons: string[]) {
+	const reasonsMap = {} as { [reason: string]: string };
+
+	for (const reason of reasons) {
+		let reasonText = null;
+		const reasonTest = reason
+			.toLowerCase()
+			.replace('_', '-')
+			.replace(' ', '-');
```
YLivay, line 48: I don't get why this is needed.

nilllzz: It's so when someone types "Off Topic" into the input for example, we can show the actual reason for "off-topic". Just wanted to prevent people being stupid and putting one of the predefined reasons as a custom reason, just with an underscore or a space.

PR #717, opened 2020-12-05
Block from post menu

nilllzz:
Adds a new entry to the post extra menu to directly block a user from a community.
Clicking that entry opens a new modal with the familiar block form, and the author of the selected post preset (and locked).

Blocking a user from there does not eject the post in the feed the menu was opened with.

You are not allowed to block yourself.

The menu entry does appear for posts made by collaborators, but when trying to submit the form a growl will appear that informs the user that a collaborator cannot be blocked from the community, and that the user has to manually remove them from the Moderators list first.

A user's block can also be overwritten by a new one by blocking them again, and the menu entry will appear for posts of users that are already blocked from the community.

### DIFF
--- a/src/app/components/forms/community/ban/block.vue
+++ b/src/app/components/forms/community/ban/block.vue
@@ -1,6 +1,8 @@
+<script lang="ts" src="./block"></script>
+
 <template>
 	<app-form name="communityBlockForm">
-		<app-form-group name="username">
+		<app-form-group v-if="!usernameLocked" name="username">
 			<app-form-control
 				type="text"
 				:rules="{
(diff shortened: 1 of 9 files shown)


src/app/components/forms/community/ban/block.vue
```diff
@@ -20,6 +22,12 @@
 				/>
 			</app-form-control-errors>
 		</app-form-group>
+		<template v-else>
+			<label class="control-label"><translate>Username</translate></label>
+			<div class="form-control -locked-username">
+				{{ formModel.username }}
+			</div>
+		</template>
```
YLivay, line 30: we should be able to just add the disabled attribute to the app-form-control yeah?
is there a specific reason you go through the trouble of not making this an app form control?

also, i know this can't happen currently, but not having an app-form-group and app-form-control-errors implies the username we initialized the modal with has to be correct since theres no longer a way to surface errors with this field through the form.

YLivay (changes requested):

src/app/components/forms/community/ban/block.vue
nilllzz: I didn't do it because I looked for a `disabled` prop on the form control and couldn't find one.
That's because there isn't one.

I didn't want to go through the trouble of adding it to form-control, cause you know, beeg control.

But I did it with a new commit, so should be allgud now.

hworld (approved):

PR #719, opened 2020-12-05
Content Viewer white-space

nilllzz:
Uses `white-space: break-spaces` (https://developer.mozilla.org/en-US/docs/Web/CSS/white-space) to render white space in Content Viewer p > span elements to match how white space is rendered in the editor.

Basically, do not collapse white space ever, and break using white space to fit the parent container restraints.

### DIFF
--- a/src/_common/content/content-viewer/content-viewer.vue
+++ b/src/_common/content/content-viewer/content-viewer.vue
@@ -17,5 +17,6 @@
 // Because white space is rendered out in the editor, we want the viewer to get as close to that
 // as possible. HTML by default collapses white space, and this overrides that behavior.
 >>> p > span
+	white-space: pre-wrap
 	white-space: break-spaces
 </style>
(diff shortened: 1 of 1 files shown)


src/_common/content/content-viewer/content-viewer.vue
```diff
@@ -13,4 +13,9 @@
 <style lang="stylus" scoped>
 >>> p > code
 	white-space: normal
+
+// Because white space is rendered out in the editor, we want the viewer to get as close to that
+// as possible. HTML by default collapses white space, and this overrides that behavior.
+>>> p > span
+	white-space: break-spaces
```
hworld, line 20: since this seems to be a little bit of a "new" property, can you use pre-wrap first and then break-spaces next since pre-wrap is the closest to break-spaces for older browsers:
```
white-space: pre-wrap
white-space: break-spaces
```

hworld (changes requested):

PR #730, opened 2021-01-01
More emojis

hacknorris-notyetbanned:
As I saw many people wanted more emojis, I created some so I've just done it. Changes : changed some source and added more graphics to /gamejolt/src/_common/emoji/ if you want, copy these emojis and put in website (◕‿◕✿)


hacknorris-notyetbanned:
ey , @hworld @YLivay why it isn't yet merged? it's here for about a month...

PR #732, opened 2021-01-04
Community icon fixes

nilllzz:
Fixes:

 - Don't shrink community avatar in selector
 - Show "Join Community" when community is not joined

### DIFF
--- a/src/app/components/shell/cbar/community/community.ts
+++ b/src/app/components/shell/cbar/community/community.ts
@@ -81,7 +81,7 @@ export default class AppShellCbarCommunity extends Vue {
 	}
 
 	get shouldShowJoin() {
-		return !this.shouldShowLeave;
+		return !this.community.is_member;
 	}
 
 	async onLeaveCommunityClick() {
--- a/src/app/components/shell/cbar/community/community.ts
+++ b/src/app/components/shell/cbar/community/community.ts
@@ -38,6 +38,7 @@ export default class AppShellCbarCommunity extends Vue {
 	@State communityStates!: Store['communityStates'];
 	@Action leaveCommunity!: Store['leaveCommunity'];
 	@Action toggleLeftPane!: Store['toggleLeftPane'];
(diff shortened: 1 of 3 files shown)


src/app/components/shell/cbar/community/community.ts
```diff
@@ -76,14 +77,23 @@ export default class AppShellCbarCommunity extends Vue {
 	}
 
 	get shouldShowLeave() {
-		return !this.community.hasPerms();
+		return !this.community.hasPerms() && !!this.community.is_member;
+	}
+
+	get shouldShowJoin() {
+		return !this.shouldShowLeave;
```
hapurubokka, line 84: We should also be checking whether or not the user is a mod/collaborator of a community here. It currently shows both `Join` and `Moderate` for the context menu of cbar items that I have collaborator status with.

src/app/components/shell/cbar/community/community.vue
```diff
@@ -1,69 +1,89 @@
...
+					<a
+						v-else-if="shouldShowJoin"
+						class="list-group-item has-icon"
+						@click="onJoinCommunityClick"
+					>
+						<app-jolticon icon="add" notice />
+						<translate>Join Community</translate>
+					</a>
+					<a
+						v-if="shouldShowModerate"
```
hapurubokka, line 74: This is currently showing both `Join` and `Moderate` for communities in which you have collaborator status. I'm not sure if we need to show both at any point, but fixing the previous issue in the TS file should resolve this just fine.

hapurubokka (changes requested):

src/app/components/shell/cbar/community/community.ts
nilllzz: `Moderate` is for backend moderation.
For a community collaborator, "Edit Community" is shown instead of "Join Community".

hapurubokka (approved):

PR #742, opened 2021-01-27
Error handler for HTTP 429 responses

nilllzz:
Implements an error handler in payload.service that handles HTTP response code 429 by showing an error growl.

### DIFF
--- a/src/_common/payload/payload-service.ts
+++ b/src/_common/payload/payload-service.ts
@@ -250,9 +250,9 @@ export class Payload {
 			this.store.commit('app/setError', 500);
 		} else if (error.type === PayloadError.ERROR_RATE_LIMIT) {
 			Growls.error({
-				title: Translate.$gettext(`Woah there, slow down!`),
+				title: Translate.$gettext(`Whoa there, slow down!`),
 				message: Translate.$gettext(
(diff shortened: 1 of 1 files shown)


src/_common/payload/payload-service.ts
```diff
@@ -234,6 +248,13 @@ export class Payload {
 			this.store.commit('app/redirect', Environment.wttfBaseUrl + '/timeout');
 		} else if (error.type === PayloadError.ERROR_INVALID) {
 			this.store.commit('app/setError', 500);
+		} else if (error.type === PayloadError.ERROR_RATE_LIMIT) {
+			Growls.error({
+				title: Translate.$gettext(`Woah there, slow down!`),
```
hworld, line 253: whoa

src/_common/payload/payload-service.ts
```diff
@@ -234,6 +248,13 @@ export class Payload {
...
+				message: Translate.$gettext(
+					`Looks like you are doing that too much. Slow down, then try again in a couple minutes.`
```
hworld, line 255: let's say "few" instead of "couple" since it's more vague in english

hworld (changes requested):

PR #755, opened 2021-03-11
Go to upload servers when uploading a file

YLivay:


src/_common/api/api.service.ts
```diff
@@ -161,7 +163,9 @@ export class Api {
 		}
 
 		const method = sanitizedPostData ? 'POST' : 'GET';
-		const url = (options.apiHost || this.apiHost) + (options.apiPath || this.apiPath) + uri;
+		const apiHost = options.apiHost || this.apiHost;
```
hworld, line 166: Change this chunk to:
```
const apiHost = options.apiHost ?? (options.file ? this.uploadHost : this.apiHost);
const url = apiHost + (options.apiPath ?? this.apiPath) + uri;
```

src/_common/api/api.service.ts
```diff
@@ -93,6 +93,7 @@ export interface RequestOptions {
 
 	// You can change the default api host/path used through these.
 	apiHost?: string;
+	uploadHost?: string;
```
hworld, line 96: Remove this option, they can just use apiHost

hworld (changes requested):

PR #778, opened 2021-04-22
Reason input remembrance

nilllzz:
Adds datalists to remember previous inputs in form fields


src/_common/settings/options.service.ts
```diff
@@ -0,0 +1,74 @@
...
+		case 'community-move-post':
+			return new Options<string>(id, true, 10);
+		case 'community-user-block':
+			return new Options<string>(id, true, 20);
+		default:
+			assertNever(type);
+	}
+}
+
+export class Options<T> {
```
YLivay, line 19: Lets call it something more specific.

src/_common/settings/options.service.ts
```diff
@@ -0,0 +1,74 @@
...
+	private _storageKey: string;
+	private _checkDuplicate: boolean;
+	private _maxItems: number | null;
+
+	constructor(id: string, checkDuplicate: boolean, maxItems: number | null) {
```
YLivay, line 24: lets get rid of checkDuplicate. I don't see where we'd even want to store duplicates in the first place

src/_common/settings/options.service.ts
```diff
@@ -0,0 +1,74 @@
...
+
+	public getList(): T[] {
+		const list = JSON.parse(this._getStorageValue());
+		if (!Array.isArray(list)) {
+			return [];
+		}
+		return list;
+	}
+
+	public setList(list: T[]) {
```
YLivay, line 50: lets make this private

src/_common/settings/options.service.ts
```diff
@@ -0,0 +1,74 @@
...
+		const value = JSON.stringify(list);
+		this._setStorageValue(value);
+	}
+
+	public unshiftItem(item: T) {
```
YLivay, line 55: I think it'd be nice to enforce some max serialized length on the item as a sanity check, just to avoid filling the localstorage. its a fairly limited resource and the data list autocomplete won't show long results in full anyways. lets try to keep this minimal and practical, esp since we might add actual templating support in the future yeah?

YLivay (changes requested):

src/_common/settings/options.service.ts
nilllzz: Well, it is inherently limited in length due to the input fields it uses. Since we control when/where we use this, we should be good with this rn.

PR #790, opened 2021-05-16
show different message for "0 views" to make it less confusing to users

nilllzz:


hworld:
Since the file changed I did this change in this commit instead https://github.com/gamejolt/gamejolt/commit/41c6ad234b4580828d2a77d97c255189079be144

PR #791, opened 2021-05-16
Fix chat autoscroll behavior

nilllzz:
Fixes the chat's autoscroll behavior, particularly with images/gifs.
These were caused because image/gif inserts would rapidly change the height of the input (because of the image loading in). These happen independent of the vue ticks, so we couldn't easily separate action frames (wait on resize -> check autoscroll).
Because of that, it would determine that the user must've scrolled, so we disabled autoscroll.
The solution is to buffer the scroll handler by 50ms so that rapid fire height changes would be caught by the same autoscroll check.

Also did some cleanup for event handling, and using emit instead of EventBus.


YLivay (commented):
not sure i like this solution. i think in general we should watch for existing elements changing their dimensions and notifying a parent observable.

PR #800, opened 2021-06-10
 Fix iOS Safari making videos fullscreen on play

hapurubokka:
- Allows Game Jolt videos to be played without going fullscreen in iOS Safari.
- Going fullscreen in iOS Safari will now open through the browser implementation.
- Adds better mute handling/syncing for iOS Safari fullscreen controls.

### DIFF
--- a/src/_common/video/player/controller.ts
+++ b/src/_common/video/player/controller.ts
@@ -30,6 +30,7 @@ export function getVideoPlayerFromSources(
 }
 
 export class VideoPlayerController {
+	muted: boolean;
 	volume: number;
 	duration = 0;
 	state: VideoPlayerState = 'paused';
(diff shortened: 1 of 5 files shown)


src/_common/video/player/controller.ts
```diff
@@ -47,6 +48,19 @@ export class VideoPlayerController {
...
+	 * iOS Safari doesn't allow us to:
+	 * - Directly change the volume of an HTMLVideoElement
+	 * - Request to make elements fullscreen in a normal way.
+	 *
+	 * If we're not able to go fullscreen our preferred way, we should be
+	 * changing the way our controls work slightly to support both muting
+	 * through external methods and going fullscreen with their controls.
+	 */
+	get altControlsBehavior() {
+		return !document.fullscreenEnabled;
```
hworld, line 61: check for GJ_IS_SSR so that you don't hit document by accidnet

src/_common/video/player/player.ts
```diff
@@ -318,7 +318,15 @@ export default class AppVideoPlayer extends Vue {
 	syncWithState() {
 		if (this.player.queuedFullScreenChange !== null) {
 			if (this.player.queuedFullScreenChange) {
-				this.$el.requestFullscreen();
+				if (this.player.altControlsBehavior) {
+					const video: any = this.$el.getElementsByTagName('video')[0];
```
hworld, line 322: I would try to make this safer just in case there is no video element for whatever reason. Maybe just use `querySelector` and check for falsey.

hworld: There's no way to do this through a vue ref, is there?

hworld (changes requested):

src/_common/video/player/player.ts
hapurubokka: We don't have access to the direct video tag here. Changed from `getElementsByTagName` to `querySelector`.

hworld (approved):

PR #802, opened 2021-06-21
Default feed

nilllzz:
- Sets "For You" as the default home route.
- Adds a setting to swap default home route back to "Following".
- Adds dot indicator on the "Following" tab for new activity.


hworld:
Did this in a different PR

PR #881, opened 2021-11-09
Content editor controller changes

hapurubokka:


src/app/views/contentedor/contentedor.route.ts
```diff
@@ -0,0 +1,7 @@
+import { RouteConfig } from 'vue-router';
```
hapurubokka, line 1: Remove these

PR #882, opened 2021-11-12
Improve fireside screen design

hapurubokka:
- Shows host list below video when desktop or portrait mobile, hiding when editing chat for mobile
- Move the fireside header into the overlay for mobile
- Move the desktop audio slider into the overlay
- Display a thumbnail of the latest video data when pausing the stream

### DIFF
--- a/src/app/components/fireside/stream/video/video.ts
+++ b/src/app/components/fireside/stream/video/video.ts
@@ -1,5 +1,5 @@
 import Vue from 'vue';
-import { Component, InjectReactive, Prop } from 'vue-property-decorator';
+import { Component, InjectReactive, Prop, Watch } from 'vue-property-decorator';
 import {
 	FiresideRTCUser,
 	FiresideVideoLock,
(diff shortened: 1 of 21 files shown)


src/app/components/fireside/stream/video/video.vue
```diff
@@ -1,11 +1,25 @@
 <script lang="ts" src="./video"></script>
 
 <template>
-	<div class="-video-player" />
+	<div class="-video-player">
+		<div ref="video" :class="{ '-hide': !shouldPlayVideo }" />
+		<canvas ref="canvas" :class="{ '-hide': shouldPlayVideo }" />
```
nilsgj, line 6: Why can't we do `v-when` instead of this? I guess you don't wanna `v-if` to not unload it.

src/app/components/fireside/stream/video/video.ts
```diff
@@ -54,4 +77,28 @@ export default class AppFiresideStreamVideo extends Vue {
...
+		if (this.pausedFrameData) {
+			const {
+				$refs: { canvas },
+				pausedFrameData: { width, height },
+			} = this;
+
+			canvas.width = width;
+			canvas.height = height;
+			const context = canvas.getContext('2d')!;
+			context.putImageData(this.pausedFrameData, 0, 0, 0, 0, width, height);
```
nilsgj, line 92: There is no way this could have transparency right? So we wouldn't need to clear what's already on the canvas?

src/app/components/fireside/stream/video/video.vue
hapurubokka: I changed it to use `v-show` on the canvas, and removed it on the video. The video hides itself already.

src/app/components/fireside/stream/video/video.ts
hapurubokka: It shouldn't, but I changed it to clear the old data just in case.

PR #891, opened 2021-12-08
Fix post form disregarding its confirmation dialog

hapurubokka:
Fixes a conflict between the `FormService` `$router.beforeEach` and the `ModalService` `$router.beforeEach`.


hapurubokka:
[When clicking back in post editing it doesn't stop you from routing away if you say not to](https://trello.com/c/cSGLtzbH/140-when-clicking-back-in-post-editing-it-doesnt-stop-you-from-routing-away-if-you-say-not-to)

src/_common/modal/AppModal.vue
```diff
@@ -54,10 +58,12 @@ onMounted(() => {
 		});
 	}
 
-	_beforeEachDeregister = router.beforeEach((_to, _from, next) => {
-		_dismissRouteChange();
-		next();
-	});
+	if (!props.ignoreRouter) {
```
hworld, line 61: Put this within the router.beforeEach so that the prop can reactively change the behavior

hworld (changes requested):

src/_common/modal/AppModal.vue
hapurubokka: Changed to use `afterEach` and dismiss only if there was no reported navigation failure. This seemed to work well during testing for both dirty post forms and dirty comment forms.

PR #892, opened 2021-12-08
Tweak fireside overlay gradient, fix some other bugs

hapurubokka:


hapurubokka:
[Change fireside controls overlay darkening to closer match something like Twitch.](https://trello.com/c/lAndKnrG/143-change-fireside-controls-overlay-darkening-to-closer-match-something-like-twitch)

hworld (approved):

PR #894, opened 2021-12-15
Split test for showing a home hero banner instead of the explore page…

hworld:
… for guests

### DIFF
--- a/src/app/components/shell/footer/footer.vue
+++ b/src/app/components/shell/footer/footer.vue
@@ -10,7 +10,7 @@
 						<br class="visible-xs" />
 						Test the beta version of the Game Jolt app.
 					</p>
-					<app-app-buttons source="footer" centered />
+					<app-app-buttons source="footer" />
 				</div>
 
 				<br />
--- a/src/app/components/shell/footer/footer.vue
+++ b/src/app/components/shell/footer/footer.vue
@@ -10,9 +10,7 @@
 						<br class="visible-xs" />
 						Test the beta version of the Game Jolt app.
 					</p>
-					<p>
-						<app-app-buttons source="footer" />
-					</p>
+					<app-app-buttons source="footer" centered />
 				</div>
 
 				<br />
--- a/src/app/views/discover/home/_fs-post/fs-post.ts
+++ b/src/app/views/discover/home/_fs-post/fs-post.ts
@@ -46,12 +46,6 @@ export default class AppHomeFsPost extends Vue {
 		return this.post.videos[0];
 	}
 
-	get bgStyle() {
-		return {
-			backgroundImage: `url('${this.mediaItem.mediaserver_url}')`,
-		};
-	}
-
 	mounted() {
 		if (this.video?.postCardVideo) {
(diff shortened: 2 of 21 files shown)


src/app/components/shell/footer/footer.vue
```diff
@@ -10,9 +10,7 @@
 						<br class="visible-xs" />
 						Test the beta version of the Game Jolt app.
 					</p>
-					<p>
-						<app-app-buttons source="footer" />
-					</p>
+					<app-app-buttons source="footer" centered />
```
hapurubokka, line 13: Is `centered` meant to be a prop, or are we assigning styling to it through this somewhere? Is this meant to be using the `justified` prop instead?

hapurubokka: This also seems to be requiring a `source!: AppPromotionSource` prop.

src/app/views/discover/home/_fs-post/fs-post.ts
```diff
@@ -0,0 +1,79 @@
...
+			backgroundImage: `url('${this.mediaItem.mediaserver_url}')`,
+		};
+	}
+
+	mounted() {
+		if (this.video?.postCardVideo) {
+			this.videoController = new VideoPlayerController(this.video.postCardVideo, null);
+
+			this.videoController.volume = 0;
+			this.videoController.muted = true;
```
hapurubokka, line 60: Would it be preferred to pass in a `VideoPlayerControllerContext` other than `null` here, either using `'gif'` or creating a different one that stays muted?

src/app/views/discover/home/_fs-post/fs-post.ts
```diff
@@ -0,0 +1,79 @@
...
+		}
+
+		return this.post.videos[0];
+	}
+
+	get bgStyle() {
+		return {
+			backgroundImage: `url('${this.mediaItem.mediaserver_url}')`,
+		};
+	}
```
hapurubokka, line 53: doesn't appear to be used

src/app/views/discover/home/home.ts
```diff
@@ -33,19 +28,45 @@ import AppDiscoverHomeCommunities from './_communities/communities.vue';
...
+					return;
+				}
+
+				// If we bootstrapped and are logged out, then we've decided to
+				// show this as a split test. If we're logged in, we always use
+				// the default, since it's used as the discover page.
+				if (!this.user) {
+					trackExperimentEngagement(configGuestHome);
+					this.split = configGuestHome.value;
+					this.split = 'hero';
```
hapurubokka, line 60: This second line should be removed, yeah?

hapurubokka (changes requested):

src/app/components/shell/footer/footer.vue
hworld: good catch, the centered was old. also, source is passed

src/app/views/discover/home/home.ts
hworld: oh dang, yeah, haha

hapurubokka (approved):

PR #919, opened 2022-02-25
Chat autoscroll improvements

hapurubokka:
- Fixes chat not always autoscrolling when performing message limit removals
- Improves moderation and history browsing by performing message limit removals only when we allow autoscrolling


src/app/components/chat/window/output/output.vue
```diff
@@ -199,10 +215,40 @@ export default class AppChatWindowOutput extends Vue {
 			return;
 		}
 
-		if (el.scrollHeight - (el.scrollTop + el.offsetHeight) > 10) {
+		const getOffsetFromBottom = () => {
```
hworld, line 218: i mean, at this point, isn't it just faster to type `el.` before each variable? haha

hworld (approved):

PR #922, opened 2022-03-26
Quests

hworld:


src/_common/fireside/post/card/AppPostCardWithText.vue
```diff
@@ -0,0 +1,229 @@
+<script lang="ts">
```
hapurubokka, line 1: @hworld do we need this? it's currently unused, and `AppPostCard.vue` has some other changes recently.

src/app/AppMain.vue
```diff
@@ -31,6 +32,16 @@ if (!import.meta.env.SSR) {
...
 
+// We allow users to access the onboarding flow even after they've gone through
+// onboarding.
+//
+// In case they did that, or didn't fully complete their onboarding, clear out
+// their onboarding-start timestamp now so they're no longer considered to be
+// onboarding.
+//
+// TODO(quests) test this
+Onboarding.clearOnboardingStartTimestamp();
```
hworld, line 43: can't be here, gotta put it in onMounted, otherwise you'll break SSR

hapurubokka: fixed 👍

PR #957, opened 2022-06-06
Fireside redesign

hapurubokka:


hworld:
there's no way this will ever merge anyway

PR #991, opened 2022-09-06
Remote content capabilities

nilsgj:
...


hapurubokka (approved):
Made a couple changes while reviewing

PR #1005, opened 2022-09-20
Allow to paginate supporters for posts and games.

hworld:


src/_common/supporters/AppSupportersModal.vue
```diff
@@ -0,0 +1,122 @@
+<script lang="ts" setup>
+import { computed, onMounted, PropType, ref, toRefs } from 'vue';
+import AppUserList from '../../app/components/user/list/AppUserList.vue';
```
hapurubokka, line 3: importing app

src/_common/user/follow/AppUserFollowButton.vue
```diff
@@ -0,0 +1,132 @@
...
+	} else {
+		try {
+			const result = await ModalConfirm.show(
+				$gettext(`Are you sure you want to unfollow this user?`),
+				$gettext(`Unfollow user?`)
+			);
+
+			if (!result) {
+				return;
+			}
```
hapurubokka, line 96: Check for the result outside of the try{}catch{}finally{} block, otherwise we'll track an unfollow when the user doesn't confirm.

hapurubokka (changes requested):

src/_common/user/follow/AppUserFollowButton.vue
hworld: wow, i didn't actually know that finally ALWAYS runs, even if you early return out from it

PR #1027, opened 2022-10-28
Update to reflect joltex changes on ChatRoomChannel and ChatRoomPresence

ramagj:


src/app/components/chat/user-collection.ts
```diff
@@ -193,38 +193,6 @@ export class ChatUserCollection {
 		return curUser;
 	}
 
-	online(input: number | ChatUser) {
```
hworld, line 196: We need to keep these functions since we use them in the ChatUserChannel. We still track in their friends list if people are online or offline, we just don't care to track for chat room users

src/app/components/chat/room-channel.ts
```diff
@@ -357,7 +339,7 @@ export async function createChatRoomChannel(
 				const user = roomMembers.get(+id) ?? new ChatUser(roomPresence.user);
 				user.typing = roomPresence.metas.some(meta => meta.typing);
 				roomMembers.update(user);
-				roomMembers.online(+id);
```
hworld, line 360: just remove, don't comment

hworld (changes requested):

src/app/components/chat/user-collection.ts
```diff
@@ -198,13 +198,11 @@ export class ChatUserCollection {
 		if (!user) {
 			return;
 		}
-
```
hworld, line 201: add newlines back in

PR #1042, opened 2022-12-03
Desktop video streaming support

hworld:


hapurubokka (approved):
I think I'd like a way to show the source selection window again without having to unset and set the `Desktop video capture` source again, but it isn't entirely required for us to test this further. Tested on Linux, everything looked great.

PR #1056, opened 2022-12-20
Chat member diffs

hworld:


src/app/components/chat/client.ts
```diff
@@ -433,43 +405,39 @@ export function setMessageEditing(chat: ChatClient, message: ChatMessage | null)
...
 
-export function retryFailedQueuedMessage(chat: ChatClient, message: ChatMessage) {
+export function retryFailedQueuedMessage(room: ChatRoom, message: ChatMessage) {
 	if (!message._isQueued || !message._error) {
 		return;
 	}
 
-	// Assign new logged on time and order queued messages by date.
 	message.logged_on = new Date();
+	arrayRemove(room.queuedMessages, i => i.id === message.id);
```
hapurubokka, line 414: Shouldn't we insert this back into the `queuedMessages` array, sorting it to be the newest message?

It looks like this current implementation would remove the message from the messages display, and our message would be lost forever if it failed to send again (or if we couldn't find a room channel for this room).

`sendChatMessage` appears to be removing it from `queuedMessages` after the message is successfully sent anyway, so I think we just need to update `logged_on` and update the index for this message.

src/app/components/chat/user-collection.ts
```diff
@@ -48,26 +52,39 @@ export class ChatUserCollection {
 		}
 	}
 
+	/**
+	 * Use this to fully replace the list of users we're tracking.
+	 */
+	replace(newUsers: ChatUser[]) {
+		this.doBatchWork(() => {
+			this.users.splice(0, Infinity, ...newUsers);
```
hapurubokka, line 60: can use the `arrayAssignAll` util

src/app/views/fireside/_sidebar/AppFiresideSidebarHosts.vue
```diff
@@ -3,15 +3,16 @@ import { computed, ref } from 'vue';
 import { stringSort } from '../../../../utils/array';
 import { sleep } from '../../../../utils/utils';
 import {
-	inviteFiresideHost,
-	removeFiresideHost,
+inviteFiresideHost,
+removeFiresideHost
```
hapurubokka, line 7: formatting

src/app/components/chat/room-channel.ts
```diff
@@ -262,22 +267,35 @@ export async function createChatRoomChannel(
...
 
-	function _onMemberLeave(data: MemberLeavePayload) {
-		const roomMembers = client.roomMembers[roomId];
+	function _onMemberInc(data: MemberIncPayload) {
+		room.value.member_count += data.amount;
+	}
+
+	function _onMemberDec(data: MemberDecPayload) {
+		room.value.member_count -= data.amount;
+	}
```
hapurubokka, line 286: Seems to be some issues with syncing these properly. Testing over a couple tabs, different users may have different counts (4 members counted instead of 2), and refreshing one of the tabs can bring a member count from 2 to 0.

hapurubokka (changes requested):

PR #1057, opened 2022-12-20
Test to show post lead content below recommended post cards

hapurubokka:


src/_common/content/content-viewer/components/AppContentViewerParagraph.vue
```diff
@@ -0,0 +1,28 @@
...
+		type: Object as PropType<ContentObject>,
+		required: true,
+	},
+});
+
+const { contentData } = toRefs(props);
+
+const owner = useContentOwnerController()!;
+
+const rootComponent = computed(() => (owner.contentRules.inlineParagraphs ? 'span' : 'p'));
```
hworld, line 18: oddly enough, I think there's some restrictions on what can go inside certain tags. span is HTML-defined inline, so it doesn't like putting block level within it. i think to be safe we should actually just use a div, or even still a p tag, and override the styling with "display: inline"
https://stackoverflow.com/questions/11314668/what-elements-can-a-span-tag-contain-in-html5

src/_common/content/content-viewer/components/AppContentViewerBaseComponent.vue
```diff
@@ -53,23 +54,54 @@ function getComponentType(data: ContentObject): any {
...
 	if (childObjects) {
+		let prevWasParagraph = false;
+		const inlineParagraphs = contentRules?.inlineParagraphs === true;
+
 		for (const contentData of childObjects) {
+			const currentIsParagraph = contentData.type === 'paragraph';
+
+			// Add a space of padding around each paragraph we're trying to
+			// render inline.
+			if (inlineParagraphs && (prevWasParagraph || currentIsParagraph)) {
```
hworld, line 71: Can you make the paragraphs inline-block and just put some margin around them? Or does it not collapse the margins properly with our HTML structure?

src/_common/content/content-viewer/components/AppContentViewerBaseComponent.vue
```diff
@@ -53,23 +54,54 @@ function getComponentType(data: ContentObject): any {
 	}
 }
 
-export function renderChildren(childObjects: ReadonlyArray<ContentObject>) {
+export function renderChildren(
```
hworld, line 57: generally I don't break out the general functions and just put everything in the script/setup block to keep things simple. it'll still be super fast since I'm pretty sure the JS engine just optimizes it anyway

src/_common/content/content-viewer/components/AppContentViewerParagraph.vue
hapurubokka: Changed to use the `p` tag with `display: inline`

src/_common/content/content-viewer/components/AppContentViewerBaseComponent.vue
hapurubokka: Inline-block doesn't work to collapse text in the way we want it here.

src/_common/content/content-viewer/components/AppContentViewerBaseComponent.vue
hapurubokka: This function is exported, which is why it's outside of the script/setup block. The helper function above it is also used within this exported function.

PR #1065, opened 2023-01-03
Show processing state when sending bulk supporter messages

hapurubokka:


src/app/views/dashboard/supporters/RouteDashSupporters.vue
```diff
@@ -102,6 +106,12 @@ const { reload } = createAppRoute({
 		hasError.value = false;
 		reachedEnd.value = rawActions.length < ACTIONS_PER_PAGE;
 	},
+	onDestroyed() {
+		if (_sendAllInterval) {
```
hworld, line 110: this piece of code is duplicated 4 times, make a cleanup function for it and call it in the 4 places

hworld (changes requested):

hworld (approved):

PR #1074, opened 2023-01-19
Fireside realms

YLivay:


src/_common/fireside/realm/realm.model.ts
```diff
@@ -0,0 +1,25 @@
+import { Model } from '../../model/model.service';
+import { Realm } from '../../realm/realm-model';
+
+export class FiresideRealm extends Model {
+	fireside_id!: number;
```
hworld, line 5: we use declare now for models like this:
```
declare fireside_id: number;
```

It tells typescript of the field without actually outputting anything. The syntax we have right now technically would output a defineAttribute for the field with typescript, but I have disabled that funcionality while we migrate over.

src/_common/fireside/realm/realm.model.ts
```diff
@@ -0,0 +1,25 @@
...
+	realm!: Realm;
+
+	added_on!: number;
+	featured_on!: number | null;
```
hworld, line 9: We usually make these optional instead and not include them in the constructor:
```
declare featured_on?: number;
```

src/_common/fireside/realm/realm.model.ts
```diff
@@ -0,0 +1,25 @@
...
+
+	constructor(data: Partial<FiresideRealm> = {}) {
+		super(data);
+
+		this.featured_on = data.featured_on ?? null;
```
hworld, line 14: remove

src/_common/realm/realm-model.ts
```diff
@@ -27,7 +27,7 @@ export class Realm extends Model {
 
 	get routeLocation(): RouteLocationDefinition {
 		return {
-			name: 'realms.view',
+			name: 'realms.view.overview',
```
hworld, line 30: use the imported route: `routeRealmsViewOverview.name`

src/app/views/realms/view/firesides/RouteRealmViewFiresides.vue
```diff
@@ -0,0 +1,106 @@
...
+	if (Screen.isXs) {
+		return 4;
+	} else if (Screen.isSm) {
+		return 5;
+	}
+
+	return 6;
+});
+
+const placeholderCount = computed(() => {
```
hworld, line 43: generally for things that just return I use the shorthand `computed(() => gridColumns.value * 2);` it'll put it in one line then

src/app/views/realms/view/firesides/RouteRealmViewFiresides.vue
```diff
@@ -0,0 +1,106 @@
...
+	// 2 rows for all breakpoints
+	return gridColumns.value * 2;
+});
+
+const appRoute = createAppRoute({
+	routeTitle: computed(() =>
+		routeStore.realm.value ? `Firesides in the ${routeStore.realm.value.name} Realm` : null
```
hworld, line 50: use `$gettextInterpolate` to make this translatable

src/app/views/realms/view/firesides/RouteRealmViewFiresides.vue
```diff
@@ -0,0 +1,106 @@
...
+
+	return 6;
+});
+
+const placeholderCount = computed(() => {
+	// 2 rows for all breakpoints
+	return gridColumns.value * 2;
+});
+
+const appRoute = createAppRoute({
```
hworld, line 48: destructure this instead: `const {isBootstrapped} = `

src/app/views/realms/view/firesides/RouteRealmViewFiresides.vue
```diff
@@ -0,0 +1,106 @@
...
+				display: 'grid',
+				gridTemplateColumns: `repeat(${gridColumns}, 1fr)`,
+				gridGap: '16px',
+			}"
+		>
+			<template v-if="!firesides.length">
+				<AppFiresideAvatarBase
+					v-for="i of placeholderCount"
+					:key="`placeholder-${i}`"
+					:is-placeholder="true"
```
hworld, line 93: use the shorthand of just `is-placeholder` without a value

src/app/views/realms/view/overview/RouteRealmsOverview.vue
```diff
@@ -0,0 +1,154 @@
...
+				url: `/web/posts/fetch/realm/${route.params.path}`,
+				shouldShowFollow: true,
+				itemsPerPage: payload.perPage,
+			},
+			payload.items,
+			fromCache
+		);
+
+		// TODO(fireside-realms) this does not exist on this payload. need a new overview endpoint.
+		firesides.value = []; // Fireside.populate(payload.activeFiresides);
```
hworld, line 84: did you make the endpoint yet or what?

src/app/views/realms/view/overview/RouteRealmsOverview.vue
```diff
@@ -0,0 +1,154 @@
...
+<template>
+	<div>
+		<div
+			:style="{
+				display: `flex`,
+				justifyContent: `space-between`,
+			}"
+		>
+			<h4 class="section-header">
+				<AppTranslate>Firesides</AppTranslate>
```
hworld, line 108: convert these to gettext

src/app/views/realms/view/overview/RouteRealmsOverview.vue
```diff
@@ -0,0 +1,154 @@
...
+			</h4>
+
+			<AppButton
+				trans
+				:to="{
+					name: routeRealmsViewFiresides.name,
+					params: { path: routeStore.realm.value.path },
+				}"
+			>
+				<AppTranslate>View All</AppTranslate>
```
hworld, line 118: Also when converting to gettext, change these from "Title Case" to "Sentence case", so "View all"

src/app/views/realms/view/overview/RouteRealmsOverview.vue
```diff
@@ -0,0 +1,154 @@
...
+const feed = ref(null) as Ref<ActivityFeedView | null>;
+const firesides = ref<Fireside[]>([]);
+const isBootstrapped = ref(false);
+
+// How many firesides to show in the preview row, including the "add a fireside".
+const firesidesGridColumns = 5;
+
+const displayablePreviewFiresides = computed(() => {
+	// -1 to leave room for the "add a fireside"
+	const perRow = firesidesGridColumns - 1;
```
hworld, line 52: this is also a const, move out of the computed

src/app/views/realms/view/overview/RouteRealmsOverview.vue
```diff
@@ -0,0 +1,154 @@
...
+	return Object.freeze(firesides.value.slice(0, perRow));
```
hworld, line 53: we don't freeze

src/app/views/realms/view/RouteRealmsView.vue
```diff
@@ -16,80 +14,39 @@ import AppShareCard from '../../../../_common/share/card/AppShareCard.vue';
...
 	}),
 };
 </script>
 
 <script lang="ts" setup>
 const router = useRouter();
-const route = useRoute();
 
-const realm = ref<Realm>();
+const routeStore = createRealmRouteStore();
```
hworld, line 33: destructure from this after setting it up, it's just nicer to call `realm` below instead of `routeStore.realm` each place

src/app/views/realms/view/RouteRealmsView.vue
```diff
@@ -99,31 +56,36 @@ function onShareClick() {
...
-		appRoute: appRoute,
-		route: route,
-		router: router,
-	});
-}
 </script>
 
 <template>
-	<AppShellPageBackdrop v-if="realm">
+	<AppShellPageBackdrop v-if="routeStore.realm.value">
```
hworld, line 62: After you destructure you won't need to use .value in the templates anymore

hworld: also remove <AppTranslate>

src/app/views/realms/view/overview/RouteRealmsOverview.vue
YLivay: thats what i asked you about with route splitting. i thought i could maybe avoid refetching realm since the parent route has it but theres no way to refresh anything else like known followers and preview firesides if only the parent realm does it.

how about we just make every route refetch the realm info and update the view store? this way we can pull everything to the sidebars once we figure out a better layout.

src/app/views/realms/view/firesides/RouteRealmViewFiresides.vue
YLivay: i dont like comments being outside the functions when they refer to a specific line within them. for one liners tho i guess its fine?

src/app/views/realms/view/RouteRealmsView.vue
YLivay: remove what?

src/app/views/fireside/_sidebar/AppFiresideSidebarFiresideSettings.vue
```diff
@@ -52,13 +55,57 @@ const {
...
+}
+
+function removeRealm(realm: Realm) {
+	const idx = selectedRealms.value.findIndex(i => i.id === realm.id);
+	if (idx === -1) {
+		console.warn('Attempted to remove a realm that is not attached');
+		return;
+	}
+
+	selectedRealms.value.splice(idx, 1);
```
hapurubokka, line 105: Change to `arrayRemove`. Can check the results of the function to do the same thing if needed.

src/app/components/forms/fireside/AppFormFiresideAdd.vue
```diff
@@ -156,31 +202,34 @@ function onRemoveCommunity() {
 			<AppFormControlErrors />
 		</AppFormGroup>
 
-		<template v-if="canSelectCommunity">
+		<template v-if="canAttachTargets">
 			<AppFormGroup
-				class="-group-community"
+				class="-group-targettables"
 				name="community_id"
 				:label="$gettext(`Start in a community?`)"
```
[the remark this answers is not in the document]
hapurubokka: `targettables` => `targetables`

src/app/views/realms/view/view.store.ts
```diff
@@ -0,0 +1,42 @@
...
+	knownFollowers: ModelClassType<User>[];
+	knownFollowerCount: number;
+	activeFiresides: ModelClassType<Fireside>[];
+};
+
+export function createRealmRouteStore() {
+	const realm = ref() as Ref<Realm>;
+	const knownFollowers = ref<User[]>([]);
+	const knownFollowerCount = ref(0);
+	const firesides = ref<Fireside[]>([]);
```
hapurubokka, line 22: I think that we should also have `userFireside` here. Community route was getting both some preview firesides along with our own fireside, so we would replace the FiresideAdd component with our own fireside if it existed. Check `src/app/views/communities/view/overview/overview.vue` for how we handled the firesides there.

src/app/views/realms/view/overview/RouteRealmsOverview.vue
```diff
@@ -0,0 +1,153 @@
...
+import { AppActivityFeedLazy } from '../../../../components/lazy';
+import AppPostAddButton from '../../../../components/post/add-button/AppPostAddButton.vue';
+import { routeRealmsViewFiresides } from '../firesides/firesides.route';
+import { useRealmRouteStore } from '../view.store';
+
+export default {
+	...defineAppRouteOptions({
+		resolver: ({ route }) =>
+			Promise.all([
+				Api.sendRequest('/web/realms/' + route.params.path),
```
hapurubokka, line 27: Do we need to request the realm again here? The parent route seems to be doing the same thing.

src/app/views/realms/view/overview/RouteRealmsOverview.vue
```diff
@@ -0,0 +1,153 @@
...
+const feed = ref(null) as Ref<ActivityFeedView | null>;
+const isBootstrapped = ref(false);
+
+// How many firesides to show in the preview row, including the "add a fireside".
+const firesidesGridColumns = 5;
+
+const displayablePreviewFiresides = computed(() => {
+	// -1 to leave room for the "add a fireside"
+	return firesides.value.slice(0, firesidesGridColumns - 1);
+});
```
hapurubokka, line 53: check `src/app/views/communities/view/overview/overview.vue` for how this was handled

src/app/views/realms/view/view.store.ts
```diff
@@ -0,0 +1,42 @@
...
+
+export type RealmRoutePayload = {
+	realm: ModelClassType<Realm>;
+	knownFollowers: ModelClassType<User>[];
+	knownFollowerCount: number;
+	activeFiresides: ModelClassType<Fireside>[];
+};
+
+export function createRealmRouteStore() {
+	const realm = ref() as Ref<Realm>;
```
hapurubokka, line 19: Is there a reason we're not defining this as `ref<Realm>()` so it's nullable? It looks like a lot of places that access this are checking for nullability anyway.

src/app/views/fireside/_sidebar/AppFiresideSidebarFiresideSettings.vue
```diff
@@ -256,6 +303,28 @@ function onClickChatMods() {
...
+									:max-communities="1"
+									:realms="selectedRealms"
+									:max-realms="maxRealms"
+									:targetable-communities="targetableCommunities"
+									can-add-realm
+									can-remove-realms
+									:with-community-channels="false"
+									@remove-realm="removeRealm"
+									@select-realm="attachRealm"
+								/>
```
hapurubokka, line 323: Targets have the same background as our sidebar, making things blend together too much. Can we have `AppPostTargets` tag a `bg-color` prop?

hapurubokka (changes requested):

src/_common/popper/AppPopper.vue
```diff
@@ -299,6 +312,8 @@ onMounted(() => {
 			}
 		},
 	});
+
+	onManualShow();
```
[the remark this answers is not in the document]
YLivay: this might not be how we're used to doing things. the watcher didnt seem to trigger when the component loaded so even if you set trigger=manual and manual-show it wouldnt show the popper when it loads in.

is there a better way to trigger this to happen?

src/_common/realm/AppRealmFullCard.vue
```diff
@@ -27,6 +27,14 @@ const props = defineProps({
...
 		default: undefined,
 	},
+	/**
+	 * By default when `to` prop is specified the entire card is linkified.
+	 * If `toLinkifiesImage` is true, it'll only linkify the image.
+	 */
+	toLinkifiesImage: {
+		type: Boolean,
+		default: false,
+	},
```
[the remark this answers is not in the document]
YLivay: didnt realize you can make more specific props like this, cool :D

src/_common/realm/AppRealmFullCard.vue
```diff
@@ -95,15 +103,14 @@ async function onClick(event: Event) {
 	<div
 		class="-card"
 		:class="{
-			'-hoverable': to || followOnClick,
```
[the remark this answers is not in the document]
YLivay: in general i think follow on click doesnt make sense even if the entire thing is linkified. i would expect to go to the realm if its a cursor point while hovering over the image, not follow the user. imo it would only make sense to expand the click target to the entire space below the image in the sheet but not more than that.

regardless, i'll revert it to what it was since that's how we had it. just think its odd.

src/_common/realm/AppRealmFullCard.vue
```diff
@@ -115,9 +122,11 @@ async function onClick(event: Event) {
 		/>
 
 		<AppResponsiveDimensions :ratio="REALM_CARD_RATIO">
-			<AppMediaItemBackdrop v-if="mediaItem" :media-item="mediaItem">
-				<AppImgResponsive class="-cover-img" :src="mediaItem.mediaserver_url" alt="" />
-			</AppMediaItemBackdrop>
+			<RouterLink v-if="to" :to="to">
```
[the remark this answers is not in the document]
YLivay: i didnt realize you could use RouterLink in `component :is`, thats neat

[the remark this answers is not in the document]
YLivay: can we use "template" instead so we don't unnecessarily nest it?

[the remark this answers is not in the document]
YLivay: doesnt seem to work. dang, can't do a "vanishing component" either because each component must have one root element, and if a component vanishes and has multiple children it'll end up rendering multiple root components either. guess its not worth it.

src/_common/realm/realm-model.ts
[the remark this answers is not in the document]
YLivay: ah, quite the conundrum. what do we do with other components? i imagine many of our Model's routeLocation functions suffer from the same limitation.

src/app/components/fireside/avatar/AppFiresideAvatar.vue
```diff
@@ -269,12 +279,25 @@ async function ejectFireside(community: FiresideCommunity) {
...
+									<AppTranslate>and</AppTranslate>
+								</template>
+							</template>
+
+							<template v-if="realm">
+								<div class="-tooltip-img-realm">
+									<AppRealmThumbnail :realm="realm" not-rounded />
+								</div>
+								{{ realm.name }}
+							</template>
```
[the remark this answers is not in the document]
YLivay: ah, i didnt actually touch these, are we fine doing these random refactors in PRs? I've been trying to avoid these to make the PR smaller and more readable.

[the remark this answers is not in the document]
YLivay: is there a better way than this mess?
```
<template v-if="i.isFeatured">
	{{ $gettext(`Unfeature fireside`) }}
</template>
<template v-else>
	{{ $gettext(`Feature fireside`) }}
</template>
```
If i remember correctly we must use constant expressions within our $gettexts but this feels horribly wordy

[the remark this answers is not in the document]
YLivay: this is not very readable to me ```{{ i.isFeatured ? $gettext(`Unfeature fireside`) : $gettext(`Feature fireside`) }}```

src/app/components/fireside/avatar/common.styl
```diff
@@ -166,12 +166,15 @@ $-zindex-popper = 4
...
 	&-members
 		font-weight: 700
 
 	&-user
-	&-community
+	&-muted
 		font-size: $font-size-small
 
-	&-community
+	&-muted
```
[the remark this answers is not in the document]
YLivay: what do we do instead?

we fine with duplication? or should we just pull these into js?

src/app/components/forms/fireside/AppFormFiresideAdd.vue
```diff
@@ -66,18 +63,20 @@ const form = createForm({
...
 		if (payload.targetableCommunities) {
-			communities.value = Community.populate(payload.targetableCommunities);
+			targetableCommunities.value = Community.populate(payload.targetableCommunities);
 		} else {
-			communities.value = [];
+			targetableCommunities.value = [];
+		}
+
+		if (payload.maxRealms) {
+			maxRealms.value = payload.maxRealms;
```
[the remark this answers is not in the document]
YLivay: we don't have any current plans to do this, but who knows. it was easy to support so didn't see why not to.

src/app/components/fireside/badge/add/add.vue
```diff
@@ -43,7 +47,11 @@ export default class AppFiresideBadgeAdd extends Vue {
 	};
 
 	async onClickBadge() {
-		const fireside = await FiresideAddModal.show({ community: this.community });
+		const fireside = await FiresideAddModal.show({
+			community: this.community,
+			realms: this.realms || [],
```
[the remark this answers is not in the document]
YLivay: do we need to do this if the prop doesnt have a default?
in this case it eventually goes to `AppFiresideAddModal` which has
```
realms: {
		type: Array as PropType<Realm[]>,
		default: () => [],
	},
```
if we didn't have the default here would have have to specify `PropType<Realm[] | undefined>`?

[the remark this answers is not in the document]
YLivay: i remember we used to have RequiredProp vs OptionalProp or something like that. do we still have these? not sure whats our current style with this.

src/app/components/forms/fireside/AppFormFiresideAdd.vue
```diff
@@ -125,16 +133,54 @@ function onBlurTitle() {
...
-function onRemoveCommunity() {
-	form.formModel.community_id = null;
+function removeRealm(realm: Realm) {
+	const idx = realms.value.findIndex(i => i.id === realm.id);
+	if (idx === -1) {
+		console.warn('Attempted to remove a realm that is not attached');
+		return;
+	}
+
+	realms.value.splice(idx, 1);
```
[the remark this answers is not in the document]
YLivay: I don't like `arrayRemove`:
1. It should return empty array if it didnt find the element to splice to have a consistent return type.
2. It's unclear whether it removes all elements or just the first one from its name.
3. Theres no way to specify what to do when the element does not exist. This is a common use case.
I made some changes to `arrayRemove` and checked all references.

src/app/components/post/AppPostTargets.vue
```diff
@@ -1,5 +1,9 @@
 <script lang="ts" setup>
+// TODO(fireside-realms) refactor this into a component thats not specific for posts.
+// Its basically a general purpose "attachable targets".
```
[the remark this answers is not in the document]
YLivay: what do you think we should call this? AppAttachableTargets? should this live in _common?

src/app/views/fireside/AppFiresideHeader.vue
```diff
@@ -115,6 +124,11 @@ const { fireside, stickerTargetController, overlay } = toRefs(props);
 	height: 16px
 	margin-right: 4px
 
+.-realm-avatar
+	width: 12px
+	height: 16px
+	margin-right: 8px
+
```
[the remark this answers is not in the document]
YLivay: that kind of calculation isnt only used here tho. its hard to tell everywhere its used too. need to look at any component that directly or indirectly displays a realm cover image which is just out of scope of this PR.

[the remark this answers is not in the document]
YLivay: yeah, i think we should do this separately. If i'll attempt doing this now i'll probably miss a bunch of places and that'll make it harder to find them later..

src/app/views/realms/view/overview/RouteRealmsOverview.vue
YLivay: i asked david about this. the issue is i want to always fetch the fresh data so i can update the sidebar rendered by the parent, but the parent doesn't bother reloading when switching between it's children. so i just had both children refetch and update a common store.

maybe we can do something like the parent deferring displaying the shell until the child route loads in and populates the store? this way we can just not fetch in the parent route.

src/app/views/realms/view/overview/RouteRealmsOverview.vue
YLivay: i did, david commented that we shouldnt use Object.freeze if thats what youre referring to. tbh I don't understand why its doing that anyways, and as for being able to create a fireside - thats not needed. currently you can always create a fireside in any realm, so always leave room for add a fireside.

its the same situation if you already have a fireside running. in this case the -1 would be leaving room for your own fireside.

src/app/views/realms/view/view.store.ts
YLivay: i think i did this when my volar was broken hah. ill check.

src/app/components/fireside/avatar/AppFiresideAvatarBase.vue
```diff
@@ -80,6 +86,10 @@ const hasLink = computed(() => {
 				/>
 			</div>
 
+			<div v-if="realm" class="-realm">
+				<AppRealmThumbnail v-if="!isPlaceholder" class="-realm-img" :realm="realm" />
```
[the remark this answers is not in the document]
YLivay: yeah, tbh i dont know how we want to display this. when i asked david about maybe placing a similarly looking rectangle with the realm on the other side he was against it. also when we allow multiple realms how would this look?

src/app/views/fireside/_sidebar/AppFiresideSidebarFiresideSettings.vue
YLivay: what would this prop accept? is it a name for a css variable? do we have scaffolding classes for applying these colors or what?

src/_common/realm/realm-model.ts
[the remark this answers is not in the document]
hworld: yeah, keep the model's using the string, and when routing within a section you can use the import

src/app/components/fireside/avatar/AppFiresideAvatar.vue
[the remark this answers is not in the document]
hworld: I think you would get used to the ternary way honestly. it's pretty readable to me. also, we talked in chat but you would do $gettextInterpolate($ngettext(...), {})

src/app/components/fireside/avatar/common.styl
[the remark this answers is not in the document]
hworld: when rewriting a whole component use the JS way, but if you're making a small addition, you just type out the whole name through duplication, yeah

src/app/components/forms/fireside/AppFormFiresideAdd.vue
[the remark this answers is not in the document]
hworld: we use arrayRemove everywhere, so we should definitely use it here

src/app/views/realms/view/view.store.ts
YLivay: hey, maybe im looking in the wrong place but i don't see it being used.
its used when you fetch the home route but doesnt seem to show on the community itself.
whats the intended behaviour for this?

src/app/components/forms/fireside/AppFormFiresideAdd.vue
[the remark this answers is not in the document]
YLivay: made the default 0, should be gud now

src/app/views/realms/view/overview/RouteRealmsOverview.vue
YLivay: clarification, this is referring to AppHomeFireside's displayFiresides basically.

fix it for both communities route and realms route

src/app/components/fireside/avatar/AppFiresideAvatarBase.vue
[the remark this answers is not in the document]
YLivay: ill put realm rectangle with outline where the community currently shows and boop the community to the right side (below the cog)

src/app/components/forms/fireside/AppFormFiresideAdd.vue
```diff
@@ -10,43 +12,39 @@ import AppFormControlErrors from '../../../../_common/form-vue/AppFormControlErr
...
 type FormModel = {
 	title: string;
-	community_id: number | null;
 	auto_feature: boolean;
 	add_community_as_cohosts: boolean;
 };
 
 const props = defineProps({
 	community: { type: Object as PropType<Community>, default: null },
+	realms: { type: Array as PropType<Realm[]>, default: () => [] },
```
hapurubokka, line 28: This list of realms is getting modified directly. If you add a realm through the selector and close the modals, any previously selected realms are still appearing when the FiresideAdd modal is opened again. Probably need to clone the list when initializing instead of using the `realms` ref everywhere.

hapurubokka (changes requested):

src/app/views/fireside/_sidebar/AppFiresideSidebarFiresideSettings.vue
YLivay: sorry, this ended up being very annoying to do, had to propogate this through 3 components all the way to pill, and pill-bi and sometimes it could just be an <a> tag and it none of this was even used by this PR.

instead i just pass in a `bg-color-offset` bool prop, and if its given it applies a class that does a deep selector on `.pill` to change its background color. its very limited yes but its local to the only component that uses it rn, so im fine with it.

when we have more usecases for this we can do this

src/app/components/forms/fireside/AppFormFiresideAdd.vue
YLivay: im not sure if this is the best way to fix this:
```
const realmsActual = ref<Realm[]>([...realms.value]);
```
after destructuring props. is this how we do things?

PR #1087, opened 2023-02-06
Vite 4, upgraded vue and some other packages as well

hworld:


src/_common/firebase/firebase.service.ts
```diff
@@ -14,11 +14,6 @@ const firebaseConfig = {
 	measurementId: Environment.firebaseMeasurementId,
 };
 
-export function initFirebase() {
-	// Just calling this should initialize it and save it.
-	getFirebaseApp();
-}
```
YLivay, line 20: is this just not used in web or has been relocated?

vite.config.ts
```diff
@@ -13,16 +13,6 @@ import { readFromViteEnv } from './scripts/build/vite-runner';
...
 
-type ViteUserConfig = ViteUserConfigActual & {
-	// This is an experimental feature, and as of time of writing
-	// does not exists with the type definition the package exports.
-	// They do work tho: https://vitejs.dev/guide/ssr.html#ssr-externals
-	ssr?: {
-		external?: (string | RegExp)[];
-		noExternal?: (string | RegExp)[];
-	};
-};
```
YLivay, line 24: yay

ssr/bundle-runner.js
```diff
@@ -40,7 +44,22 @@ module.exports.compileModule = filepath => {
...
+				// file.
+				if (id.startsWith('.')) {
+					id = path.resolve(path.dirname(filepath), id);
+				}
+				return require(id);
+			},
+			m,
+			filepath,
+			path.dirname(filepath)
+		);
```
YLivay, line 62: comment on why the require function has problems with relative paths in the first place and if theres some documentation you can point at for this. i don't fully understand why this fixes ssr.

vite.config.ts
```diff
@@ -325,6 +317,7 @@ export default defineConfig(async () => {
 						// traffic from port 443 to 8443 in a separate process.
 						// Example of using socat for that is in the README.md
 						port: process.platform === 'darwin' ? 8443 : 443,
+						host: true,
```
YLivay, line 320: add a comment on why that was needed and how its fixing it.

vite.config.ts
```diff
@@ -409,8 +402,8 @@ export default defineConfig(async () => {
 
 					// For the mobile app build, we currently can't load
 					// cross-origin requests, so we want to essentially make
-					// just one big JS chunk.
-					if (gjOpts.platform === 'mobile') {
+					// just one big JS chunk. SSR also works better this way.
+					if (['mobile', 'ssr'].includes(gjOpts.platform)) {
```
YLivay, line 406: what does removing ssr here do?

src/_common/bootstrap.ts
```diff
@@ -28,7 +28,7 @@ export type BootstrapOptions<T = Component> = {
  * point can call to get what it needs.
  */
 export async function bootstrapCommon(options: BootstrapOptions) {
-	const router = options.router ?? null;
+	const router = options?.router ?? null;
```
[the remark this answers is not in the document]
hworld: oh, this was the proper thing to do based on the typing actually, so i just fixed it when i saw it

src/_common/firebase/firebase.service.ts
hworld: not used at all

vite.config.ts
hworld: SSR builds now were outputting multiple files, which was screwing things up, so I'm putting it back to how we had it now where it just outputs a single file like the mobile app build does. I'm sure I can get this working if I spend enough time, but I don't want to waste time doing all these minor things

PR #1095, opened 2023-02-16
Fix badge positioning in outline buttons

hworld:


src/_common/button/button.styl
```diff
@@ -58,20 +45,27 @@ make-button-size(size)
...
-		theme-prop('border-color', 'bi-bg', true)
+		change-bg(
+			'bi-bg',
+			true
+		)
+		theme-prop(
+			'border-color',
+			'bi-bg',
+			true
+		)
```
hapurubokka, line 56: Any idea what changed to cause this to happen in our `.styl` files now?

hapurubokka (approved):

src/_common/button/button.styl
hworld: upgraded stylus recently, maybe it's that. or just the stylus extension we used updated

PR #1098, opened 2023-02-21
Group chat firesides

nilsgj:


src/app/components/grid/firesides-channel.ts
```diff
@@ -0,0 +1,36 @@
...
+	const { socketController, appStore } = client;
+
+	const logger = createLogger('Fireside');
+
+	const channelController = createSocketChannelController('firesides', socketController);
+
+	channelController.listenTo('expired', _onFiresideExpired);
+
+	const joinPromise = channelController.join({
+		async onJoin() {},
```
hapurubokka, line 22: Do we need to do anything during `onJoin`? We can just remove that object if we don't need to do anything when joining.

src/app/store/index.ts
```diff
@@ -86,6 +88,8 @@ export function createAppStore({
 	const communities = ref<Community[]>([]);
 	const communityStates = ref(new CommunityStates());
 
+	const firesides = ref<StoredFireside[]>([]);
```
hapurubokka, line 91: While we add/remove/update things in this list, it doesn't actually seem to be used for anything. Should we remove this and the related functions, or was there something else planned for this?

hapurubokka (changes requested):

src/app/store/index.ts
nilsgj: Ah yeah, I wanted to implement a global Firesides channel that one can subscribe to for fireside expiration updates. I reverted those changes in this branch.

hapurubokka (approved):

PR #1100, opened 2023-02-22
Sort dev games

nilsgj:


src/_common/game/game.model.ts
```diff
@@ -516,3 +519,24 @@ export async function unfollowGame(game: Game) {
...
+			break;
+
+		default:
+			showErrorGrowl($gettext(`You cannot add a new game or collaborate on this game.`));
+			router.push({
+				name: 'library.collection.developer',
+				params: { id: user.username },
+			});
+	}
+}
```
hapurubokka, line 542: We can probably just put the router push at the end of this function since we'll always call it.

hapurubokka (approved):

PR #1104, opened 2023-02-24
Allow post to have both backgrounds and media

hapurubokka:


src/app/components/forms/post/AppFormPost.vue
```diff
@@ -182,9 +177,11 @@ const form: FormController<FormPostModel> = createForm({
 			enableVideo();
 		} else if (_model.hasMedia) {
 			enableImages();
-		} else if (attachmentType.value !== '') {
+		} else if (attachmentType.value) {
 			enabledAttachments.value = true;
-		} else if (form.formModel.background) {
+		}
+
+		if (form.formModel.background) {
```
hapurubokka, line 184: Check background_id pref if it exists

hworld (changes requested):
do the things

src/app/components/forms/post/AppFormPost.vue
hapurubokka: We're doing this during `onLoad` of the form instead of `onInit`

PR #1109, opened 2023-03-05
Update Notification Text

TheCrafters001:
The notification currently shows `User (@user) is live!.`, which has bugged me for some time. Here is a fix to remove the period after the exclamation mark. The notification should now show `User (@user) is live!`

### DIFF
--- a/src/_common/notification/notification-text.service.ts
+++ b/src/_common/notification/notification-text.service.ts
@@ -1,7 +1,7 @@
 import { Community } from '../community/community.model';
 import {
 	CommunityUserNotification,
-	NotificationType
+	NotificationType,
 } from '../community/user-notification/user-notification.model';
 import { formatCurrency } from '../filters/currency';
 import { FiresideCommunity } from '../fireside/community/community.model';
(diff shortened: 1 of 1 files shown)


src/_common/notification/notification-text.service.ts
```diff
@@ -1,7 +1,7 @@
 import { Community } from '../community/community.model';
 import {
 	CommunityUserNotification,
-	NotificationType,
```
nilsgj, line 4: Ensure the formatting is in line with the rest of the repository here: We use tailing commas for imports.

src/_common/notification/notification-text.service.ts
```diff
@@ -467,7 +467,7 @@ export class NotificationText {
 				if (notification.action_model instanceof Fireside) {
 					return _process(
 						$gettextInterpolate(
-							`<em>%{ subject }</em> is live!.`,
```
nilsgj, line 470: The same mistake is present in line ~507 in this file, can you fix that too?

src/_common/notification/notification-text.service.ts
nilsgj: It's probably some formatting extension not set up on your editor, either editorconfig, prettier or eslint.

nilsgj (approved):

PR #1149, opened 2023-06-16
Show notification blip on Backpack cbar item for new users

hapurubokka:


hworld (changes requested):
Move the setting of the watermark into onboarding so that it can be done in one place in app

hworld (approved):

PR #1155, opened 2023-06-26
Move quests into cbar and overlay windows, add hash redirects

hapurubokka:


src/app/components/shell/sidebar/AppShellSidebar.vue
```diff
@@ -9,6 +9,7 @@ import AppShellSidebarBackpack from './AppShellSidebarBackpack.vue';
 import AppShellSidebarContext from './AppShellSidebarContext.vue';
 import AppShellSidebarLibrary from './AppShellSidebarLibrary.vue';
 import AppShellSidebarMobile from './AppShellSidebarMobile.vue';
+import AppShellSidebarQuests from './AppShellSidebarQuests.vue';
```
hworld, line 12: switch this sidebar to be async loaded so that it doesn't have to load the code until they switch.
actually, can you also do for library and backpack? I didn't realize we're not separating that code into different chunks

src/app/components/shell/AppShell.vue
```diff
@@ -11,9 +11,11 @@ import { useSidebarStore } from '../../../_common/sidebar/sidebar.store';
 import AppStickerLayer from '../../../_common/sticker/layer/AppStickerLayer.vue';
 import { closeStickerDrawer, useStickerStore } from '../../../_common/sticker/sticker-store';
 import { useBannerStore } from '../../store/banner';
-import { useAppStore } from '../../store/index';
+import { TogglableLeftPane, useAppStore } from '../../store/index';
+import { useQuestStore } from '../../store/quest';
 import { AppClientShell, AppClientStatusBar } from '../client/safe-exports';
 import { useGridStore } from '../grid/grid-store';
+import AppQuestWindow from '../quest/window/AppQuestWindow.vue';
```
hworld, line 18: async load this, like the chat window.

src/app/components/shell/AppShell.vue
```diff
@@ -217,6 +313,16 @@ html, body
 		top: $shell-top-nav-height
 		left: calc(var(--shell-content-sidebar-width) + var(--shell-cbar-width))
 
+.quest-window
```
hworld, line 316: can we make chat-window and quest-window use the same styling so that less code loads?

src/app/components/shell/AppShell.vue
```diff
@@ -60,45 +71,124 @@ const ssrShouldShowSidebar = computed(
...
 		`--shell-content-sidebar-padding: ${sidebarPadding}px`,
-		`--shell-content-sidebar-width-base: ${sidebarWidthBase}px`,
-		`--shell-content-sidebar-width: ${sidebarWidthBase + sidebarPadding * 2}px`,
+		`--shell-content-sidebar-width-base: ${sidebarWidthBase + sidebarPadding * 2}px`,
+		`--shell-content-sidebar-width-lg: ${sidebarWidthLg + sidebarPadding * 2}px`,
+		`--shell-content-sidebar-width: ${sidebarWidth.value + sidebarPadding * 2}px`,
 	];
 });
 
+type HashEventResult = false | { sidebar: TogglableLeftPane | undefined };
```
hworld, line 104: let's go over this stuff in person so that I make sure I fully understand it

src/app/components/quest/AppQuestTimer.vue
```diff
@@ -36,31 +48,39 @@ onUnmounted(() => {
...
+	<span v-else-if="endsOn" v-app-tooltip.touchable="readableTimeExact" class="AppQuestTimer">
 		<AppJolticon class="small text-muted" icon="clock" />
 		<span class="text-muted">
-			{{ ' ' + readableTime }}
+			{{ ' ' + readableTimeRough }}
 		</span>
 	</span>
 </template>
+
+<style lang="stylus" scoped>
```
hworld, line 83: why not put in style attribute?

src/app/store/quest.ts
```diff
@@ -28,9 +59,65 @@ export function createQuestStore({
 	const dailyQuests = computed(() =>
 		_allQuests.value.filter(i => i.repeat_type === QuestRepeatType.daily)
 	);
-	const quests = computed(() =>
-		_allQuests.value.filter(i => i.repeat_type !== QuestRepeatType.daily)
-	);
+
+	const questChunks = computed(() => {
```
hworld, line 63: i want to be conscious of the amount of code we have to load in for the shell. let's see if we can maybe move this into other places to only load on demand

src/app/views/quests/view/view.route.ts
```diff
@@ -1,7 +0,0 @@
-import { RouteRecordRaw } from 'vue-router';
-
-export const routeQuestsView: RouteRecordRaw = {
```
hworld, line 3: do we have a redirect rule for these? we have links to at least the quests log before

hworld (changes requested):

hworld (approved):

PR #1172, opened 2023-07-19
Add more text to the expiry time of shop products

hapurubokka:


src/app/components/vending-machine/modal/_product/AppVendingMachineProduct.vue
```diff
@@ -74,6 +75,25 @@ const anchorStyles = computed<CSSProperties>(() => {
...
+	zIndex: overlayTagZIndex,
+};
+
+const readableEndsOn = computed(() => {
+	const endsOn = shopProduct.value.ends_on;
+	if (!endsOn) {
+		return;
+	}
+
+	return shorthandReadableTime(endsOn, {
```
hworld, line 90: I know this wasn't done as part of this PR, but the way to do translations properly is a bit harder. For trailingText, you don't know where the time part should be in a language. English would put it before, but others might do it differently. So you'd have to make it a function instead that gets passed the time as a var and then interpolate it in, like this:

`trailingText: (time) => $gettextInterpolate('%{ time } left to purchase', { time })`

src/app/components/vending-machine/modal/_purchase-modal/AppPurchaseShopProductModal.vue
```diff
@@ -170,6 +172,54 @@ const balanceRefs = { coinBalance, joltbuxBalance };
...
+	timeRemaining.value = $gettextInterpolate(`Limited time to purchase: %{ time }`, {
+		time: shorthandReadableTime(endsOn, {
+			nowText: '0s',
+			precision: 'exact',
+			allowFuture: true,
+		}),
+	});
+}
+
+watch(
```
hworld, line 197: Have you tried making this a watchEffect instead? Might simplify the code a bit.

hworld (changes requested):

hworld (approved):

PR #1188, opened 2023-09-07
Convert static Modal::show() functions

ramagj:
Convert static Modal::show() functions into exported nonstatic standalone function


ramagj:
Need to verify usage of static variable in `src/_common/content/content-editor/modals/gif/gif-modal.service.ts`

hworld (approved):

PR #1202, opened 2023-10-19
Shop sections

hapurubokka:


hworld (changes requested):
Two things:
- Let's make it so that the shop user in the modal is not part of the title but rather the body so that it scrolls away and gives more viewport height to the items.
- Better handling for guests. When you click to get joltbux, it should take them to the join page.

PR #1240, opened 2023-12-08
Add stripes to the profile shop button

hapurubokka:


hworld:
Deciding not to do this and instead just keep the highlight border.

PR #1275, opened 2025-10-17
Fixed a minor error in the linux_64 sort

RepositoryLover:
It's pretty self-explanatory if you look at the sorts of windows and macos and their 64 counterparts


hworld:
Nice catch, thanks @RepositoryLover

